PR #13, opened 2021-08-16
to add vendor folder

sdminonne:
Goal of this PR is to add vendor folder which is considered a golang best practice to track dependencies for security and _disconnected build_ (if needed).


sahare (approved):
Looks good, Thanks!

PR #14, opened 2021-08-16
Adding unit tests for backup utility functions

sahare:

Latest coverage:

Screen Shot 2021-08-16 at 9 48 46 AM

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -31,7 +31,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	backupv1beta1 "github.com/open-cluster-management-io/cluster-backup-operator/api/v1beta1"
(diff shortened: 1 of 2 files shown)


controllers/suite_test.go
```diff
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	backupv1beta1 "github.com/open-cluster-management-io/cluster-backup-operator/api/v1beta1"
+	restorev1beta1 "github.com/open-cluster-management-io/cluster-backup-operator/api/v1beta1"
```
sdminonne, line 34: Reason for this?

sahare: trying to test if adding restorev1beta1 into the scheme works with no error

sdminonne: yeah but the test should be functional: my understanding is that testing the idempotency of scheme adding (even with different import alias) should not stay in our operator.

sdminonne (approved):
LGTM

PR #15, opened 2021-08-16
Activate short test

jnpacker:
Signed-off-by: Joshua Packer <jpacker@redhat.com>

* Disables the generator call in `make test` so that prow can call `make unit-tests`
* Activates a `make unit-tests`


sahare (approved):
Looks good, Thanks!

PR #17, opened 2021-08-16
To revamp restore controller

sdminonne:
This PR contains new version of restore-controller. Main differences:

1. OnwerReference utilization
2. Change in APIs
3. EventRecorder

For the APIs modification, @jnpacker and all other reviewers:

~~restore.spec.backupname~~ `restore.spec.velerobackupname`
~~restore.status.phase~~ since  the same information is contained in `restore.status.restoreProxyReference.status.phase`
~~restore.spec.veleroConfigRestoreProxy.namespace~~ -> `restore.spec.veleroConfigRestoreProxy.veleroNamespace` since the kubernetes type system looks having a problem with `Namespace` field. See line 77 in `restore_types.go`


sdminonne:
@jnpacker, as soon you'll have a look at APIs I'll modify title.

controllers/restore_controller.go
```diff
@@ -60,92 +63,108 @@ func (r *RestoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			restoreLogger.Error(err, "unable to create velero restore for restore %s/%s", req.Namespace, req.Name)
+
+			return ctrl.Result{}, err
+		}
+		r.Recorder.Event(restore, v1.EventTypeNormal, "Velero Restore created:", fmt.Sprintf("%s/%s", veleroRestore.Namespace, veleroRestore.Name))
+		restore.Status.VeleroRestore = veleroRestore.DeepCopy()
+
+	case len(veleroRestoreList.Items) == 1:
+		if isRestoreFinsihed(restore) {
+			r.managedClustersHandler(restore)
```
jnpacker, line 96: I'm a little confused about this call, as the method defined below just returns a (true, err)

controllers/restore_test.go
```diff
@@ -25,43 +32,51 @@ var _ = Describe("Restore", func() {
...
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      restoreName,
-					Namespace: "",
-				},
-				Spec: v1beta1.RestoreSpec{
-					BackupName: "the-latest-backup",
-				},
-			}
+		})
+		/*
```
jnpacker, line 60: I'm never a big fan of commenting out code. If you don't need it, just remove it, you can go back to the previous commit if you need to bring it back, unless you intend to use it later. 😄

jnpacker:
/lgtm

* I added two small comments, not required to complete the merge if your pressed for time.

controllers/restore_controller.go
sdminonne: Sure, I'm writing this function in another PR: at the moment. Goal of this WIP was to have only a review of REST APIs.

controllers/restore_test.go
sdminonne: Right! It's a cruft: see for example [this](https://github.com/identitatem/dex-operator/pull/18#discussion_r690918118) .

sdminonne:
@jnpacker thanks for the review. I'm going to remove the commented code (sorry for that I didn't double-checked all the code before submitting PR or at least not completely).

controllers/restore.go
```diff
@@ -17,24 +17,161 @@ limitations under the License.
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	ocinfrav1 "github.com/openshift/api/config/v1"
+
 	v1beta1 "github.com/open-cluster-management/cluster-backup-operator/api/v1beta1"
+	veleroapi "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 )
 
-// IsRestoreFinsihed returns true when Restore is finished
+// isRestoreFinsihed returns true when Restore is finished
```
birsanv, line 38: typo on isRestoreFinsihed :)

controllers/restore.go
```diff
@@ -17,24 +17,161 @@ limitations under the License.
...
+
+const adminKubeconfigSecretFmt string = dns1123SubdomainFmt + "(-admin-kubeconfig)"
+
+var adminKubeconfigSecretRegex = regexp.MustCompile("^" + adminKubeconfigSecretFmt + "$")
+
+const boostrapSATokenFmnt string = dns1123SubdomainFmt + "-bootstrap-sa-token" + "([-a-z0-9]*[a-z0-9])?"
+
+var boostrapSATokenRegex = regexp.MustCompile("^" + boostrapSATokenFmnt + "$")
+
+// filterSecrets filters a list of secrets getting
```
birsanv, line 64: The function's comment looks incomplete here

birsanv (approved):
lgtm

sdminonne:
@jnpacke @barraudl  @sahare : REST APis changed (phase and messages replaced with Conditions)
PTAL

README.md
```diff
@@ -188,3 +188,16 @@ schedule-acm
 The restore operation should be run on a different hub then the one where the backup was created.
 
 After you create a `restore.cluster.open-cluster-management.io` resource on the new hub, you should be able to run `oc get restore -n <oadp-operator-ns>` and get the status of the restore operation. You should also be able to verify on your new hub that the backed up resources contained by the backup file have been created.
+
+To restore a backup you can select a specific backup specifying the name of the backup inside `veleroBackupName.spec.restore` field or you can leave it blank. If you leave it blank the opeator will select the most recent backup without errors.
```
birsanv, line 192: Dario, a small typo : opeator

Also, as we use the schedulebackup resource now, replacing the cluster backup, the name of the veleroBackup would be something like below ( sample taken from the Usage section https://github.com/open-cluster-management/cluster-backup-operator#usage )

```
kind: Restore
metadata:
  name: restore-acm
spec:
  backupName: acm-managed-clusters-schedule-20210902205438
```

controllers/restore_controller.go
```diff
@@ -19,133 +19,370 @@ package controllers
...
 var (
-	restoreRequeueInterval = time.Minute * 1
+	restoreOwnerKey = ".metadata.controller"
+	apiGVStr        = v1beta1.GroupVersion.String()
 )
 
+const (
+	managedClusterImportInterval            = 10 * time.Second // as soon restore is finished we start to poll for managedcluster registration
+	BootstrapHubKubeconfigSecretName        = "bootstrap-hub-kubeconfig"
+	OpenClusterManagementAgentNamespaceName = "open-cluster-management-agent" // TODO: double check whether this can change
```
birsanv, line 56: The ACM controller can be installed in a namespace different then the `open-cluster-management` which is the default; in this case the restore is not going to work

The backup operator finds the ACM install namespace by looking for the charts-v1 channel, which is created in the acm ns
I am not sure if this is the right approach but charts-v1 seems to be a constant for now

controllers/restore_controller.go
```diff
@@ -19,133 +19,370 @@ package controllers
...
+					Reason:  v1beta1.RestoreReasonRunning,
+					Message: fmt.Sprintf("Velero Restore %s still running", veleroRestore.Name),
+				})
+		}
+
+	default:
+		// TODO: handles multiple velero restores:
+		// check if one velero is still running... update status and wait
+		// if all finished handleManagedClusters
+		//for i := range veleroRestoreList.Items {}
```
birsanv, line 162: maybe remove this line

birsanv (approved):
lgtm Dario; a few comments but they can be handled in other tasks

README.md
sdminonne: Thanks

controllers/restore_controller.go
sdminonne: Is this valid for the managed cluster as well?

controllers/restore_controller.go
sdminonne: Sure, it's there for me. But you're right

controllers/restore_controller.go
birsanv: No, this is working for the hubs only .. where acm is installed

sdminonne: OK, thanks but this one is the `managedclsuter ACM` namespace. Need to find a way to retrieve this info. Thanks for rising up this

birsanv: got it .. maybe we can have an optional used property to define the ACM namespace ( as we did with the velero namespace on the VeleroConfigBackupProxy ); if not defined is open-cluster-management

sdminonne: @birsanv I found _the right solution_ for this. The information is stored in the cluster scoped `klusterlet.spec.namespace`.  So this is going to be ~~`OpenClusterManagementAgentNamespaceName`~~ will become `DefaultOpenClusterManagementAgentNamespaceName`. Most of the time the default is kept, so in order to avoid an useless remote REST call I'll try the default and eventually I'll get the non default info through a REST call. Many thanks to @qiujian16  for the info.

sdminonne:
@birsanv @jnpacker thanks. I'll address @birsanv comments and after the fix I'll remove the `hold` `label`. In case you want me to change something else, please advise.

birsanv:
@sdminonne let's check this in, any other comments can be addressed in a separate task

sdminonne:
@birsanv merged

PR #19, opened 2021-08-18
to fix operator-sdk issue

sdminonne:
Currently `operator-sdk`  or `kubebuilder`, whatever you use,  is lost due to mismatch between PROJECT file and the rest of codebase.
If we want to implement a `webhook` we need this to be fixed.


birsanv (approved):
lgtm

PR #20, opened 2021-08-18
implement multiple files backup

birsanv:

Changes 
- update contributing file with PR sign off details
- updated readme
- backup operator generates now 3 files ( remote cluster ns, credentials, other resources like app, policy )
   - updated backup resource status to have a list of velero backups
   - run reconcile every 10 sec when a backup is active; change the reconcile rate to the refresh time interval set in the backup resource after that
   - changed default refresh time interval set in the backup sample resource to be 60 ( minutes )
   - updated backup status to include details for all 3 backup files 
   - duration info in the backup status ( screen cap below showing 4 min ) represents the cumulative time for all 3 files
   - next backup execution runs after specified time interval, but adding to the the duration of the last backup; so if the time interval is set to 60min and the last backup took 5min, next one starts after 65min
   - backup status is set to InProgress if any of the 3 backups are still executing
   
   
Screen Shot 2021-08-18 at 5 52 22 PM

   
Screen Shot 2021-08-18 at 11 47 27 AM


controllers/backup.go
```diff
@@ -78,55 +82,333 @@ func (r *BackupReconciler) getActiveBackupName(backup *v1beta1.Backup, c client.
...
+		if backups[i].Status.Phase == "PartiallyFailed" {
+			return backups[i].Status.Phase // show PartiallyFailed second
+		}
+	}
+
+	//if here, return the state of the first backup
+	return backups[0].Status.Phase
+}
+
+// return the execution time for the last backup in this list of backups ( creds, res, cluster)
```
sdminonne, line 382: Guess `execution time` is `completion time`... Nit: I would change the comment

controllers/backup.go
```diff
@@ -78,55 +82,333 @@ func (r *BackupReconciler) getActiveBackupName(backup *v1beta1.Backup, c client.
...
+func getLastBackupCompletionTime(backup *v1beta1.Backup) *v1.Time {
+
+	if backup.Status.VeleroBackups == nil || len(backup.Status.VeleroBackups) <= 0 {
+		// no previous completed backup, can start one now
+		return nil
+	}
+
+	lastBackup := backup.Status.VeleroBackups[2]
```
sdminonne, line 390: Can this always happen to have a backup at position [2]?

controllers/utils.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// name used by the velero backup resource, created by the backup acm controller
+func getVeleroBackupName(backupName, backupNamesapce string) string {
+	return backupName + "-" + getFormattedTimeCRD(time.Now())
+}
+
+func isBackupPhaseFinished(phase v1beta1.StatusPhase) bool {
+
+	return phase == "Completed" ||
```
sdminonne, line 70: Use `const` defineid in API

controllers/backup.go
```diff
@@ -78,55 +82,333 @@ func (r *BackupReconciler) getActiveBackupName(backup *v1beta1.Backup, c client.
...
+	if err := c.List(ctx, &channels, &client.ListOptions{}); err != nil {
+		// if NotFound error
+		if !k8serr.IsNotFound(err) {
+			backupLogger.Info("channel resources NOT FOUND")
+		} else {
+			backupLogger.Error(err, "failed to get chnv1.ChannelList")
+		}
+	} else {
+		for i := range channels.Items {
+			if channels.Items[i].Name == "charts-v1" {
```
sdminonne, line 255: I need help here. Is `charts-v1` something that will never change or be configurable externally?

sdminonne (changes requested):
@birsanv I did my best to give (hopefully) reasonable feedback. This PR is pretty big and I think I'm losing some logic. May you try to have a look at my comments and eventually address or explain to me what I missed?

sdminonne (commented):
I would consider modifying the REST APIs instead of having an array of velero backups in the Status I would split the 3 resources.

birsanv:
@sdminonne thank you for the review and comments, I'll get back to it as soon as possible

controllers/backup.go
```diff
@@ -78,55 +82,333 @@ func (r *BackupReconciler) getActiveBackupName(backup *v1beta1.Backup, c client.
...
+	veleroResIdentity := types.NamespacedName{
+		Namespace: veleroBackup.Namespace,
+		Name:      fmt.Sprintf("%s-%s", veleroBackup.Name, resourceType),
+	}
+
+	veleroResBackup := &veleroapi.Backup{}
+	veleroResBackup.Name = veleroResIdentity.Name
+	veleroResBackup.Namespace = veleroResIdentity.Namespace
+
+	errBackup := r.Get(ctx, veleroResIdentity, veleroResBackup)
```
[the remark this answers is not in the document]
birsanv: fixed

controllers/backup_controller.go
```diff
@@ -95,9 +86,15 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		backup.Status.Phase = "ERROR"
```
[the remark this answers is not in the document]
birsanv: fixed

controllers/backup_controller.go
```diff
@@ -37,13 +34,9 @@ import (
...
 var (
-	backupOwnerKey      = ".metadata.controller"
-	apiGV               = "v1beta1" //v1beta1.GroupVersion.String()
-	requeueInterval     = time.Minute * 1
-	acmNS               = "open-cluster-management"
-	acmChannel          = "charts-v1"
-	backupNamespacesACM = [...]string{"open-cluster-management-agent", "open-cluster-management-hub", "hive", "openshift-operator-lifecycle-manager"}
-	backupNamespacesObs = [...]string{"open-cluster-management-observability"}
+	backupOwnerKey  = ".metadata.controller"
+	apiGV           = "v1beta1" //v1beta1.GroupVersion.String()
```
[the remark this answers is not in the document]
birsanv: fixed

api/v1beta1/backup_types.go
```diff
@@ -81,9 +81,9 @@ type BackupStatus struct {
 	// Message on the last operation
 	// +kubebuilder:validation:Optional
 	LastMessage string `json:"lastMessage"`
-	// Velero operation status
+	// Velero Backups operation status
 	// +kubebuilder:validation:Optional
-	VeleroBackup *veleroapi.Backup `json:"veleroBackup,omitempty"`
+	VeleroBackups []*veleroapi.Backup `json:"veleroBackup,omitempty"`
```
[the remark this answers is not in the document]
birsanv: fixed s; 
The reason I went with an array instead of separate resources was that I wasn't sure if we need to add more backup files to the list. It was easier to work with a generic list then to create 3 types of resources

controllers/backup_controller.go
```diff
@@ -156,23 +151,38 @@ func (r *BackupReconciler) submitAcmBackupSettings(ctx context.Context, backup *
 			// clean up old backups if they exceed the maxCount number
 			r.cleanupBackups(ctx, backup, c)
 
-			// set ACM backup configuration
-			setBackupInfo(ctx, veleroBackup, c)
+			// set ACM backup configuration for managed clusters
+			setManagedClustersBackupInfo(ctx, veleroBackup, c)
+			err = c.Create(ctx, veleroBackup, &client.CreateOptions{})
+
```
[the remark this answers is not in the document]
birsanv: I check for error when I verify if the resource exists
err := r.Get(ctx, veleroIdentity, veleroBackup)

If the resource doesn't exist, I am setting up the backup info ( this is what setManagedClustersBackupInfo does ) then I try to create the resource 
err = c.Create(ctx, veleroBackup, &client.CreateOptions{})

controllers/backup.go
```diff
@@ -78,55 +82,333 @@ func (r *BackupReconciler) getActiveBackupName(backup *v1beta1.Backup, c client.
...
+func getBackupPhase(backups []*veleroapi.Backup) veleroapi.BackupPhase {
+
+	if backups == nil || len(backups) <= 0 {
+		return ""
+	}
+
+	// get all backups and check status for each
+	for i := 0; i < len(backups); i++ {
+
+		if backups[i].Status.Phase == "InProgress" {
```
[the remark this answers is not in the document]
birsanv: fixed

controllers/backup.go
birsanv: fixed

controllers/backup.go
birsanv: changed with backup.Status.VeleroBackups[len(backup.Status.VeleroBackups)-1]

controllers/utils.go
birsanv: fixed :)

controllers/backup.go
birsanv: yes, we expect this does not change; rather then assume the acm install ns is open-cluster-management, I can use this channel name, used by installer to find the ns name
We've used this channel channel name with the app as well

birsanv:
Dario, on your comment 
>> I would consider modifying the REST APIs instead of having an array of velero backups in the Status I would split the 3 resources.

As discussed, I preferred to use a generic list instead of listing the resources one by one
I can make this change in a separate PR although the Schedule might affect the way we store and use backup status ( if we use the velero Schedule, then we don't have control over the backups so this code will be obsolete anyway )

birsanv:
@sdminonne your comments were addressed with  the last update

sdminonne (approved):
@birsanv thanks!

PR #22, opened 2021-08-30
move GH domain to open-cluster-management/

sdminonne:
Hi all, I'm proposing to fix which basically it's a `sed s/open-cluster-management-io/open-cluster-management/g`. I'm not 100% positive about this but the rule (if any) it's not crystal clear (to me).
May you have a look and (if rejected) explain the rationale? Thanks

### DIFF
--- a/controllers/backup.go
+++ b/controllers/backup.go
@@ -22,8 +22,8 @@ import (
 	"sort"
 	"time"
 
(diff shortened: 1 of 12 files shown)


birsanv (approved):
LGTM, thanks Dario

PR #29, opened 2021-09-01
update COMPONENT NAME & VERSION

jnpacker:
Signed-off-by: Joshua Packer <jpacker@redhat.com>
* Change COMPONENT_NAME AND COMPONENT_VERSION for build manifest.


birsanv (approved):
should we change the project name to match this component name ?

https://github.com/open-cluster-management/cluster-backup-operator

jnpacker:
If we're going never going to do the operator, which it doesn't look like, we probably should.

PR #33, opened 2021-09-02
to regenerate role file for channels

sdminonne:
Signed-off-by: Salvatore Dario Minonne <sminonne@redhat.com>


birsanv:
@sdminonne the channel is in this api group `apps.open-cluster-management.io`, not `cluster.open-cluster-management.io`

There is a rule already for that in the rule.yaml

https://github.com/open-cluster-management/cluster-backup-operator/blob/main/config/rbac/role.yaml#L12

PR #34, opened 2021-09-02
make backups TTL optional

birsanv:

- On Schedule resource, make veleroTTL optional ( backups never expire if option not set )
- On the backupSchedule sample yaml, replace schedule name, frpm `backup-schedule-acm` to `schedule-acm` ( name is a bit too long after adding backup type prefix )
Schedule names are now
Screen Shot 2021-09-02 at 4 40 57 PM


sahare (approved):
lgtm

PR #35, opened 2021-09-03
add maxBackups to backupschedule operator

birsanv:

Changes:
- add required property maxBackups to backupschedule resource
```
kind: BackupSchedule
metadata:
  name: schedule-acm
spec:
  maxBackups: 10 # maximum number of backups after which old backups should be removed
  veleroSchedule: 0 */6 * * * # Create a backup every 6 hours
  veleroTtl: 72h # deletes scheduled backups after 72h; optional, backups never expire if option not set
```

- updated readme to add the backupschedule resource and remove references to the backup ( we will keep only the backupschedule )

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Cluster Back up and Restore Operator.
   - [License](#license)
   - [Getting Started](#getting-started)
   - [Design](#design)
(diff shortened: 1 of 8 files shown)


sahare (approved):
Looks good, Thanks!

PR #39, opened 2021-09-09
add debug details when schedule has no status

birsanv:

If the velero pod is in error or the Velero resource was not created, schedules have no status

Add extra details for troubleshooting


sahare (approved):
lgtm

PR #41, opened 2021-09-13
don't create backup schedule if storage location is not available

birsanv:

for defect https://github.com/open-cluster-management/backlog/issues/16065


sahare (approved):
Looks good, Thanks!

PR #44, opened 2021-09-13
fix sonar vulnerabilities

birsanv:

https://github.com/open-cluster-management/backlog/issues/15943

- Fix gosec:G601 vulnerability based on 
https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop

- Fix gosec:G101 

```
When a specific false positive has been identified and verified as safe, you may wish to suppress only that single rule (or a specific set of rules) within a section of code, while continuing to scan for other problems. To do this, you can list the rule(s) to be suppressed within the #nosec annotation, e.g: /* #nosec G401 */ or // #nosec G201 G202 G203
```

### DIFF
--- a/controllers/restore.go
+++ b/controllers/restore.go
@@ -62,8 +62,8 @@ func NewManagedClusterHandler(ctx context.Context,
 		errors                                        = []error{}
 		err                      error
 	)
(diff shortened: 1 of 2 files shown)


sahare (approved):
Looks good, Thanks!

PR #49, opened 2021-09-15
use hive admin-kubeconfig secrets and restore label to filter ns

birsanv:
https://github.com/open-cluster-management/backlog/issues/16122

- in the attachManagedClusters to get the list of managed cluster namespaces : 
Instead of the current approach, where you get all ns with the label `cluster.open-cluster-management.io/managedCluster`

Get all hive kube admin secrets with the following labels
```
hive.openshift.io/secret-type=kubeconfig << this gives all admin-kubeconfig secrets
velero.io/restore-name=the-restore-file-name << this filters all above secrets and keeps only the ones created by the current restore op
```

Then run the attachManagedCluster only on the namespaces for those secrets ( and make sure the ns is processed only once, when we have multiple versions of the secret - ie applying the restore multiple times )

What we get with this approach:
- ignore from start managed cluster namespaces that don't have the hive secret
- run the attach command only on managed clusters imported with the current restore file ( reimporting the same file would try to re-attach those remote clusters, which I think it's a thing we want )

Screen Shot 2021-09-15 at 1 41 42 PM

### DIFF
--- a/controllers/restore_controller.go
+++ b/controllers/restore_controller.go
@@ -33,7 +33,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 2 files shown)


controllers/restore_controller.go
```diff
@@ -268,20 +266,36 @@ func (r *RestoreReconciler) initVeleroRestore(ctx context.Context, restore *v1be
...
 		}
-		attached, err := r.attachManagedCluster(ctx, restore, namespaceList.Items[i].Name)
+
+		// ignore already processed namespaces, assuming we have more then one kube admin secret in the same ns
+		_, ok := find(proccessedClusters, kubeAdminSecrets.Items[i].Namespace)
+		if ok {
+			// this namespace was already processed, ignore it
+			continue
+		}
+		proccessedClusters = append(proccessedClusters, kubeAdminSecrets.Items[i].Namespace)
```
sdminonne, line 296: I would implement differently, two things here:
1. We can set a namespaces has been processed as soon we succesfully run `attachManagedCluster` so the mechanism to _tag_ a namespace `processed` should go AFTER trying to `attachManagedCluster`.
2. The mechanism to _tag_ a cluster has `attached` to the new HUB should be persisted in the `Status` of the restore. I would say throuhg a `Condition`.

sdminonne (changes requested):
This is a good idea but I would change the implementation: tried to explain the reason in the review and here:
In general `attachManagedCluster` has not been written to work in `one-shot` and multiple pass is needed to `attach` a cluster: 
1. Replacing the Kubeconfig
2. Approve the CSR
3. Accept the cluster
Especially between  step 1 and step 2 we may have a long delay (even minutes) to wait and the `attachManagedClusters` cannot wait. We've multiple namespaces to attach.

controllers/restore_controller.go
birsanv: Dario, let's put this on the back burner until we get the ManagedCluster CR into the backup
We'll decide then how to set the labels and where; in that case we can probably make use of the ManagedCluster resource

PR #50, opened 2021-09-16
Update restore controller to restore all backup types

sahare:

https://github.com/open-cluster-management/backlog/issues/16185

Replaced `VeleroBackupName` in the restore spec with followings:
- VeleroManagedClustersBackupName
- VeleroResourcesBackupName
- VeleroCredentialsBackupName

All optional, for any of the three backup types, if the name is not provided by user, the operator will try to select the latest related backup with less error, to find which velero backup is related to the backup type, the backup name must contain the following hard coded schedule names:

https://github.com/open-cluster-management/cluster-backup-operator/blob/a6825ad294d4207530a933059954d897c8bbdae4/controllers/schedule_controller.go#L62-L66

The operator will skip the restore if can't find a proper backup name for the type, and will continue with the next type


birsanv (approved):
lgtm

PR #51, opened 2021-09-17
add managed clusters obj to backup

birsanv:
Signed-off-by: Valentina Birsan <vbirsan@redhat.com>

With the attached changes , after backing up the managed cluster resources ( cluster level and ns ) I can see the hive cluster being attached when I restore on a new hub

Backup content :
- Namespaces : hive, openshift-operator-lifecycle-manager, <managed clusters ns> , **<cluster pools ns .. this is new >** 
- Resources: 
```
    - secret
    - configmap
    - ManagedCluster.cluster.open-cluster-management.io
    - ManagedClusterAddon
    - ServiceAccount
    - ManagedClusterInfo
    - ManagedClusterSet
    - ManagedClusterSetBindings
    - KlusterletAddonConfig
    - ManagedClusterView
    - ManagedCluster.clusterview.open-cluster-management.io
    - ClusterPool
    - ClusterDeployment
    - MachinePool
    - ClusterProvision
    - ClusterState
    - ClusterSyncLease
    - ClusterSync
```

The cluster not available on hub a during restore ( they were in hibernating state ) show as pending import
Screen Shot 2021-09-17 at 11 35 23 AM
t

PR #53, opened 2021-09-20
to get rid of post-processing while restoring

sdminonne:
Since after https://github.com/open-cluster-management/cluster-backup-operator/pull/51 there's no more need for restore post-processing this PR remove attachManagedClusters and all generated rbacs.


birsanv (approved):
lgtm thanks!

PR #58, opened 2021-09-21
check only existing backups created by the acm schedule

birsanv:

related to https://github.com/open-cluster-management/backlog/issues/16310

Changes:
- when getting the list of old backups, filter in only the ones generated by the acm scheduler; backups with names


sahare (approved):
lgtm, thanks!

PR #60, opened 2021-09-23
restore op can be executed once

birsanv:

Update readme to document that restore op can be executed only once

fyi @thuyn-581 

https://github.com/open-cluster-management/backlog/issues/16170


sahare (approved):
Looks good, thanks!

PR #64, opened 2021-09-27
check if restore resource is in the velero ns

birsanv:

https://github.com/open-cluster-management/backlog/issues/16526

Changes:
- verify if the velero storageLocation object is in the same ns with the backup or restore operator resources
- show an error if they are not in the same namespace ; in the log and the resource status
- updated restore tests


sahare (approved):
Looks good, Thanks!

PR #67, opened 2021-10-05
To replace Conditions with phase and message

sdminonne:
To fix https://github.com/open-cluster-management/backlog/issues/16561


controllers/restore_controller.go
```diff
@@ -91,75 +90,40 @@ func (r *RestoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		restore.Status.Phase = v1beta1.RestorePhaseError
+		restore.Status.LastMessage = msg
 
 		// retry after failureInterval
 		return ctrl.Result{RequeueAfter: failureInterval}, errors.Wrap(
 			r.Client.Status().Update(ctx, restore),
-			updateStatusFailedMsg,
-		)
-	}
-
```
sahare, line 108: Is removing the code below intentional? If so, any reason for that not being required?

sahare (changes requested):
Thanks @sdminonne , looks good, the unit tests failed so may need a retest or fix, and a block of code was removed and I wasn't sure why :)

controllers/restore_controller.go
sdminonne: Yeah, it is intentional. The previous code was looking for the veleroStorageLocations but it was taking the first in list. (line 110). I found that one unit-tests were passing because it was selecting a veleroStorageLocations in another namespace while a veleroStorageLocations is available in the same namespace.

sdminonne: At line 104 (green block) I've added a similar logic. In case I misuderstood the current logic, I'm happy to review the code. Thanks

sdminonne: Instead I need sometime to investigate the unit tests failure. Thanks!

birsanv: Dario, the intended logic for that code was to block the restore operation to start if the storage location resource was not created or was not in Available state. We then use this StorageLocation object's namespace as the oadp operator namespace. 
We check later on if the namespace where the restore object was created matches the StorageLocation namespace; it should .. 
You found the flaw, if we have more than one storage location resources, the first one is selected and used.
I am not sure if we can have the perfect solution here; nothing can stop you to create more storages, in any ns, with a a status of available.
To make sure we get the StorageLocation used by the konveyor resource can we ( suggestion )
- get the list of the storagelocations
- get only the resources with an onwerReference not null and Kind Velero


I just checked, with oadp 0.3 you can create only one Velero resource so that would give you only one storage location

birsanv: But I suggest to have this checked in the way it is now so we can test the Phase changes
Tomorrow is the last day of sprint 5, we want this in the build by then :)

birsanv (approved):
lgtm

PR #69, opened 2021-10-06
Check if BackupStorageLocation is valid

sahare:

1. Check if BackupStoargeLocation is valid based on the logic suggested here: https://github.com/open-cluster-management/cluster-backup-operator/pull/67#discussion_r722448653

  - get the list of the storagelocations
  - get only the resources with an onwerReference not null and Kind Velero

2.  Make sure the resource is created in the namespace of the valid BackupStorageLocation


birsanv (approved):
lgtm
Sahar, we need the same logic for the backup controller..

PR #85, opened 2021-11-23
update graphic for readme

birsanv:

Changes:
- update readme image with the one available on rhacm doc
- Update error path message to remove check for not found elements.
When the list is empty client.List doesn't return an error so that case doesn't go through the error path


sahare (approved):
lgtm

PR #91, opened 2021-12-10
refactor backup content

birsanv:

https://github.com/open-cluster-management/backlog/issues/18199

Related to 
https://issues.redhat.com/projects/ACM/issues/ACM-1001

Changes: 
- updated managed cluster backup to include only resources used by backup; as tested , only this is needed to backup remote clusters : 
   	= to show the clusters as detached ( not try to connect to the new hub )
        "ClusterDeployment",
        "MachinePool",
        "ClusterSyncLease",
        "ClusterSync",
Secrets with this label: "hive.openshift.io/secret-type"

	= to activate the clusters ( activation data )

        "ManagedCluster.cluster.open-cluster-management.io", //global
        "KlusterletAddonConfig",
        "ManagedClusterAddon",
        "ManagedClusterInfo",
        "ManagedClusterSet",
        "ManagedClusterSetBindings",
        "ClusterPool",
        "ClusterCurator",
        "ManagedCluster.clusterview.open-cluster-management.io",
        "ManagedClusterView",

- with the above, I split the managed cluster backup in 2 : passive data and activation data
	= passive data , the 4 CRs , go under the resources backup; so when you restore the resources backup + secrets, you see all managed clusters in detached state. Original hub is still in charge, active; so we can move this data to multiple clusters and prepare them to take over during a disaster recovery. In this way the recovery time is very fast, we only need to restore the activation backup.
	= activation stays under the managed cluster backup; so when you restore this backup , the hub becomes the active one

- credentials backup have now 2 more backups : hive and cluster credentials
This is because we can’t include secrets in a backup and have an OR clause ( secrets have hive.openshift.io/secret-type OR cluster.open-cluster-management.io/type OR cluster.open-cluster-management.io/backup )

### DIFF
--- a/config/samples/cluster_v1beta1_restore_passive.yaml
+++ b/config/samples/cluster_v1beta1_restore_passive.yaml
@@ -1,0 +1,11 @@
+# apply this resource to restore hub resources in a passive configuration
+# restored managed clusters will show as detached on the current hub, where this resource is applied
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: Restore
+metadata:
+  name: restore-acm-passive
+  namespace: openshift-adp
+spec:
+  veleroManagedClustersBackupName: skip
+  veleroCredentialsBackupName: latest
+  veleroResourcesBackupName: latest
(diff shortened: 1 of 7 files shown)


sahare (approved):
lgtm, thanks!

PR #92, opened 2021-12-10
fix gosec G101 false positive

birsanv:

https://github.com/securego/gosec#annotating-code


sahare (approved):
lgtm

PR #96, opened 2021-12-17
create generic resources backup

birsanv:
https://github.com/open-cluster-management/backlog/issues/18200

Changes:

- Create a new resources backup for resources with label `cluster.open-cluster-management.io/backup`
```
spec:
  defaultVolumesToRestic: false
  excludedResources:
    - secret
  hooks: {}
  includeClusterResources: true
  labelSelector:
    matchExpressions:
      - key: cluster.open-cluster-management.io/backup
        operator: Exists
```
- sort schedule names to create first the credentials schedules, then clusters, last resources

- updated the tests to use len(veleroScheduleNames) instead of 3-6

- updated this tests - acm backupschedules confused with velero schedules
```
			// count ACM ( NOT velero) schedules, should be still just 5
			// this nb is NOT the nb of velero backup schedules created from
			// the acm backupschdule object ( which is len(acmSchedulesList.Items))
			// it represents the nb of ACM - BackupSchedule.cluster.open-cluster-management.io - schedules created by the tests
			acmSchedulesList := v1beta1.BackupScheduleList{}
			Eventually(func() bool {
				err := k8sClient.List(ctx, &acmSchedulesList, &client.ListOptions{})
				return err == nil
			}, timeout, interval).Should(BeTrue())
			Expect(len(acmSchedulesList.Items)).To(BeNumerically("==", 5))
```

### DIFF
--- a/config/samples/cluster_v1beta1_backupschedule.yaml
+++ b/config/samples/cluster_v1beta1_backupschedule.yaml
@@ -1,8 +1,15 @@
+#  veleroTtl - optional; deletes scheduled backups after specified time
+#  if veleroTtl is not specified, the maximum default value set by velero is used - 720h
+# 
+# veleroSchedule - cron job to start new acm backups
+#
+# maxBackups - maximum number of acm backups after which old acm backups should be removed
+#
 apiVersion: cluster.open-cluster-management.io/v1beta1
 kind: BackupSchedule
 metadata:
   name: schedule-acm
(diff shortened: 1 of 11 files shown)


sahare (approved):
Looks good, just put on hold for unit test investigations, Thanks.

sahare (approved):
Looks good, Thanks!

PR #106, opened 2022-01-07
fix delete old backups

birsanv:

https://github.com/stolostron/backlog/issues/18821

Changes : 
- use backups creationTime to find related backups; related backups should be in the 2s time interval on creation
- they also should have the prefix as defined by the acm backups list


sahare (approved):
lgtm

PR #129, opened 2022-02-08
clean up resources before restore

birsanv:

Add an optional spec property on restore to allow cleaning up resources before restoring again

`cleanupBeforeRestore` - optional, if not define is set to false ( need more testing before being confident this is working in all cases and not breaking anything )

If cleanupBeforeRestore = true, for each restore file , prior to creating the velero.io.Restore ( so calling the actual restore ) we call a routine which cleans up all resources defined by the restore's backup definition ; resources must have the label `velero.io/backup-name` so they should have been created by another restore. All other resources are kept as is

We need to test resources with finalizers or owned by other resources
For example ClusterDeployment - which I think I tested owns hive secrets and has a finalizer.
I run the secrets restore last to allow them to recreate hive secrets if changed

```
kind: Restore
metadata:
  name: restore-acm-passive-activate
spec:
  cleanupBeforeRestore: false. <<<<
  veleroManagedClustersBackupName: latest
  veleroCredentialsBackupName: skip
  veleroResourcesBackupName: skip
```

### DIFF
--- a/api/v1beta1/restore_types.go
+++ b/api/v1beta1/restore_types.go
@@ -58,6 +58,14 @@ type RestoreSpec struct {
 	// backup_name points to the name of the backup to be restored
 	// +kubebuilder:validation:Required
 	VeleroCredentialsBackupName *string `json:"veleroCredentialsBackupName"`
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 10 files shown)


sahare (approved):
Looks good, Thanks!

PR #132, opened 2022-02-09
Ignore restore if other restores are running

sahare:

https://github.com/stolostron/backlog/issues/19537


birsanv (approved):
lgtm thanks!

PR #134, opened 2022-02-10
delete resources after patch

birsanv:
Changes : 
- call clean up for all restores and after that call the restore
- call patch after delete when deleted resource uses finalizers


sahare (approved):
lgtm

PR #141, opened 2022-02-11
clean up generic resources

birsanv:
Changes: 

- Before restore is called clean up any resource with a label `cluster.open-cluster-management.io/backup` and `velero.io/backup-name` ( created by another restore )
- exclude `local-cluster` NS from managed cluster resources
- use api group for backupManagedClusterResources list


sahare (approved):
lgtm

PR #142, opened 2022-02-11
enable cleanup resources by default

birsanv:

update samples to clean up previously restored resources 

`veleroManagedClustersBackupName: true` asks to delete all resources with a velero-name label , which are resources created by a previous backup
We want to clean up before restoring again
This should address resources updated after the previous backup was run or resources delete since then


sahare (approved):
lgtm

PR #144, opened 2022-02-14
label backups with cluster id and check for collisions

birsanv:

Changes 
1. introduce a new backup status `SchedulePhaseBackupCollision`
2. annotate all backup schedules with the current cluster ID ( use `ClusterVersion` uid ) 
```
	// BackupScheduleClusterUIDLabel is the label key used to identify the cluster id that generated the backup
	BackupScheduleClusterLabel string = "cluster.open-cluster-management.io/backup-cluster"

```

sample labels 

```
 labels:
    cluster.open-cluster-management.io/backup-cluster: 86941453-3823-4f03-8bdd-082297d32129 <<<<<<
    cluster.open-cluster-management.io/backup-schedule-name: schedule-acm
    cluster.open-cluster-management.io/backup-schedule-type: credentialsCluster

```

3. when creating a new acm backupscehdule, label velero io schedules with the above label
4. on reconcile, if the acm backupschedule exists ( so NOT when creating a new schedule because in this case we assume the user specifically wants to backup from this cluster) then check if the latest resources backup is from the same cluster. If not, set the backupschedule Phase to SchedulePhaseBackupCollision

Let the backups still go through; I am going to update the policy to alert if the backupschedule Phase is BackupCollision

```
oc get backupschedule -A                                
NAMESPACE       NAME           PHASE             MESSAGE
openshift-adp   schedule-acm   BackupCollision   Backup acm-resources-schedule-20220214161948, from cluster with id [d93d09b6-3ca1-449f-9c30-dcccf2848485] is using the same storage location. Collision with this cluster backup!
```

5. Reconcile every hour to look for new collisions , using `collisionControlInterval`

### DIFF
--- a/api/v1beta1/schedule_types.go
+++ b/api/v1beta1/schedule_types.go
@@ -41,6 +41,9 @@ const (
 	// SchedulePhaseUnknown means the schedule has been processed by
 	// the ScheduleController but there are some unknown issues
 	SchedulePhaseUnknown SchedulePhase = "Unknown"
+	// another cluster pushes backups to the same storage location
+	// resulting in a backup collision
+	SchedulePhaseBackupCollision SchedulePhase = "BackupCollision"
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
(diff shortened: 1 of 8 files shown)


sahare (approved):
lgtm

PR #147, opened 2022-02-15
delete schedules on collision

birsanv:

If a backup finds a collision ( there is another schedule, from another cluster, that created more recent backups ) than the current backup will delete its schedules so that it doesn't write over the other backups.
The admin should verify which cluster must be the valid primary writing backups and decide which schedule to use


sahare (approved):
lgtm

PR #148, opened 2022-02-15
allow schedule to get through when phase is empty

birsanv:

When the backupschedule is first created, the state could be empty or New
For these 2 cases don't check on collision with other backups


sahare (approved):
lgtm

PR #152, opened 2022-02-16
fix delete on restore

birsanv:

make the cleanupOption to default deleting only resources created by restore


sahare (approved):
Looks good, Thanks!

PR #153, opened 2022-02-17
fix code smells

birsanv:

changes : 
- split function which is too long
- reduce nb of parameters for function

https://github.com/stolostron/backlog/issues/19928


sahare (approved):
lgtm

PR #163, opened 2022-03-02
backup collision fixes

birsanv:
Updates
- when checking for backup collision check if the velero schedules for this backup were created in the last few minutes, in this case skip the collision validation, this is a new backup created by the user
- update readme 
- update backup sample ttl to 3days; the cron jobs runs every 2h so more then 3 days will get too many backups
- when deleting resources on cleanup log patching finalizers; run finalizers only if delete has no errors


sahare (approved):
lgtm

PR #165, opened 2022-03-03
create a dummy backup for policy validation of enabled backups

birsanv:

https://github.com/stolostron/backlog/issues/19776

Design: 
- create a dummy backup which sets the TTL to be the backup schedule cron job + 5 min
- this dummy backup (acm-validation-policy-schedule) will delete itself 5 min after the owner backup would start a new set of backups
- if the owner backup no longer creates backups, then after the last deletion there will be no acm-validation-policy-schedule backup in the storage location
- if there are no backupschedules writing backups at that storage location then there are going to be no acm-validation-policy-schedule backups ( they delete themself using the cron time )
- the policy checks for at least one instance of a acm-validation-policy-schedule backup at the storage location; if none found, it will throw an exception , it means that there are no primary hubs backing up data as expected

Example : 
Create this backup schedule 

```
kind: BackupSchedule
metadata:
  name: schedule-acm-1
  namespace: openshift-adp
spec:
  veleroSchedule: 0 */2 * * *
  veleroTtl: 72h
```

This triggers the creation of the 6 backups
There is going to be a 7' schedule, for the acm-validation-policy-schedule
This one is set to 

```
apiVersion: velero.io/v1
kind: Schedule
metadata:
  name: acm-validation-policy-schedule
  namespace: openshift-adp
spec:
  schedule: 0 */2 * * *
  ttl: 2h <<< instead of 27h, this backup expires as soon as a new cron job should run
```

### DIFF
--- a/controllers/backup.go
+++ b/controllers/backup.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
+	"github.com/robfig/cron/v3"
 	v1beta1 "github.com/stolostron/cluster-backup-operator/api/v1beta1"
 	veleroapi "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


sahare (approved):
Looks good, Thanks!

PR #168, opened 2022-03-08
Fix the index of backup schedule names in the swap after sorting them

sahare:

The index of backup schedule names were changed after ValidationSchedule was added


birsanv (approved):
lgtm

PR #169, opened 2022-03-08
generic resources restored in activation phase

birsanv:
https://github.com/stolostron/backlog/issues/20486

backup should allow third party to backup generic resources under activation data

Changes :

- introduced a new value for for the backup label to allow restroring resources in the activation phase
when you add the `cluster.open-cluster-management.io/backup` label to a resource, this resource is automatically backed up under the `acm-resources-generic-schedule` backup. If any of these resources need to be restored only when the managed clusters are moved to the new hub, so when the `veleroManagedClustersBackupName:latest` is used on the restored resource, then you have to set the label value to `cluster-activation`. This will ensure the resource is not restored unless the managed cluster activation is called.
Example :
```yaml
apiVersion: my.group/v1alpha1
kind: MyResource
metadata:
  labels:
    cluster.open-cluster-management.io/backup: cluster-activation
``` 
- remove RequeueAfter: collisionControlInterval when backupschedule is in collision phase
- used fully qualified names for excludedCRDs resources ( use apigroup ); easier to debug and see what resources is owned by what apigroup

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Cluster Back up and Restore Operator.
   - [Getting Started](#getting-started)
   - [Design](#design)
     - [What is backed up](#what-is-backed-up)
+      - [Resources restored at managed clusters activation time](#resources-restored-at-managed-clusters-activation-time)
     - [Scheduling a cluster backup](#scheduling-a-cluster-backup)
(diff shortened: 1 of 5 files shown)


sahare (approved):
Looks good, Thanks!

PR #175, opened 2022-03-11
Update backup status to collision before deleting Velero schedules

sahare:
https://github.com/stolostron/backlog/issues/20557


birsanv (approved):
lgtm

PR #194, opened 2022-03-29
exclude local-cluster managedcluster from backup

birsanv:

related to https://github.com/stolostron/backlog/issues/21054

Until the velero.io/exclude-from-backup=true is set on the local-cluster by the installer team, I am adding the managed-cluster resource to the prepareForBackup routine and add this label there
I will take out the code as soon as the label is added when the cluster is created

This is to fix this restore warning ( at the moment velero doesn't attempt to update existing resources, which is why this is no t affecting the restore procedure; when velero will attempt to update existing resources with restore content, which is post 2.5, the local-cluster resource would be updated with the restore content, and the restore cluster is messed up )
`"warnings":{"cluster":["could not restore, managedclusters.cluster.open-cluster-management.io \"local-cluster\" already exists. Warning: the in-cluster version is different than the backed-up version."]
`

### DIFF
--- a/controllers/schedule.go
+++ b/controllers/schedule.go
@@ -27,6 +27,8 @@ import (
 	v1beta1 "github.com/stolostron/cluster-backup-operator/api/v1beta1"
 	veleroapi "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


birsanv:
fixed by installer https://github.com/stolostron/multiclusterhub-operator/pull/774

PR #215, opened 2022-04-05
Fix the phase update when sync is set to disabled

sahare:

Fixes the status phase update when syncRestoreWithNewBackups is updated from true to false, in this case the sync should stop and the phase should update from Enabled to Finished/FinishedWithErrors


birsanv (approved):
lgtm

PR #222, opened 2022-04-07
Fixes for restore sync

sahare:

https://github.com/stolostron/backlog/issues/21478

- Fix sync to check if either new resources backup or new credential backup is available to sync with (used to look for new resources backup only)
- Fix creating Velero restore code not to return error in case the resource already exists (may happen in sync case only)
- Fix cleanup code to skip cleanup if Velero restore already exists and skip ResourcesGeneric cleanup in sync activation case (fix for duplicate cleanups in activation process)

Screen Shot 2022-04-07 at 1 35 04 PM
Screen Shot 2022-04-07 at 2 24 46 PM

### DIFF
--- a/controllers/restore.go
+++ b/controllers/restore.go
@@ -171,8 +171,9 @@ func deleteDynamicResource(
 		return false, false
 	}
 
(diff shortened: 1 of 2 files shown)


birsanv (approved):
lgtm ; thanks!

PR #228, opened 2022-04-13
backup clustersync with the managed clusters backup

birsanv:

https://github.com/stolostron/backlog/issues/21018

Issue : 

Most of the hive controllers will ignore a CD with no associated ClusterSync object
There's no ClusterSync object initially because it's a hiveinternal CR and the backup has been ignoring those (which should be the right thing theoretically / long-term, but read on...)
The ClusterSync controller refuses to create a ClusterSync object for an unreachable cluster.

Fix: 
Add ClusterSync to the backup data so that hibernating clusters CD are being processed by the hive controllers


sahare (approved):
lgtm

PR #232, opened 2022-04-26
exclude policies with policy.open-cluster-management.io/root-policy label

birsanv:
https://github.com/stolostron/backlog/issues/21946


sahare (approved):
Looks good, Thanks!

PR #234, opened 2022-04-26
skip root policies on cleanup

birsanv:

When cleaning up resources before restore, skip resources with label `policy.open-cluster-management.io/root-policy` since those are not backed up

https://github.com/stolostron/backlog/issues/21946


sahare (approved):
lgtm

PR #241, opened 2022-05-09
backup clustersets and bindings under resources backup

birsanv:

related to https://github.com/stolostron/cluster-backup-operator/pull/239

https://github.com/stolostron/backlog/issues/22236

While testing the argo app placed on local cluster, we noticed that the managed cluster set ns placement binding was not saved for the default managed cluster set.
The fix with  https://github.com/stolostron/cluster-backup-operator/pull/239 was to always delete managed cluster sets even if not created by a previous restore, to allow the default managed cluster set placement binding to be recreated
It turns out thought that the ns placement binding is a different resource, which would be stored by the backup and recreated; the fact was that this placement binding resources, together with the managed cluster set are created by the managed clusters backup, so when the hub becomes active. There is no reason to not backup these resources on a passive cluster, so doing this with this PR. 

Changes:
- update readme limitation typo
- update testcases for includedResources backup; different type of backup use different type of resources
- remove the code cleaning up managed cluster set on all cases
- move the following under the resources backup; managed cluster sets and set bindings will be created on a passive hub, not only when the hub becomes active

```
"managedclusterset.cluster.open-cluster-management.io",
		"managedclusterset.clusterview.open-cluster-management.io",
		"managedclustersetbinding.cluster.open-cluster-management.io",
```

Screen Shot 2022-05-09 at 11 26 45 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ These limits work well with regular scenarios but may need to be updated when yo
 ```
   limits:
     cpu: "2"
-    memory: 512Mi
+    memory: 1Gi
   requests:
     cpu: 500m
     memory: 256Mi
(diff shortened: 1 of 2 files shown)


sahare (approved):
Looks good, Thanks!

PR #243, opened 2022-05-11
resources with backup label should always be backed up by the generic backup

birsanv:

With the exception of the secrets and configmaps which are backed up by the credentials schedule, all other resources having the `cluster.open-cluster-management.io/backup-cluster` should be backed up under the acm-resources-generic-schedule schedule

https://github.com/stolostron/backlog/issues/22478


sahare (approved):
lgtm thanks

PR #258, opened 2022-06-30
use backupSchedule ttl for MSA validity

birsanv:

https://github.com/stolostron/backlog/issues/23876

Changes : 
- if backupschedule ttl time is set, use that for MSA validity; MSA validity = ttl * 3
- if backupschedule not set, use default 720h * 3


sahare (approved):
lgtm

PR #260, opened 2022-07-04
Fix for prepareForBackup call

sahare:

https://github.com/stolostron/backlog/issues/24012


birsanv (approved):
lgtm

PR #262, opened 2022-07-05
Post restore activation

sahare:
https://github.com/stolostron/backlog/issues/23681


birsanv:
@sahare the secret name is `auto-import-account`, not `auto-import-secret`

birsanv (approved):
lgtm

PR #265, opened 2022-07-06
add spec prop to enable MSA

birsanv:

https://github.com/stolostron/backlog/issues/24033

Backup schedule has 2 new optional properties used to enable MSA
- useManagedServiceAccount if set to true the MSA account is generated
- managedServiceAccountTTL if specified, used for the token expiration time; if not set the time is veleroTTL*2
- have to update hub-crds repo with the new CRD

Sample:
```
kind: BackupSchedule
metadata:
  name: schedule-acm
spec:
  veleroSchedule: 0 */2 * * *
  veleroTtl: 120h
  useManagedServiceAccount: true
```


sahare (approved):
lgtm

PR #276, opened 2022-07-25
update encryption info

birsanv:

Recommend to use server side encryption


sahare (approved):
lgtm

PR #301, opened 2022-09-06
Set human readable time format for logger

sahare:

https://github.com/stolostron/backlog/issues/25372


birsanv (approved):
lgtm; thanks!

birsanv:
@sahare we also need this cherrypicked in 2.6 so it gets picked up by the 2.6.1 stream

PR #309, opened 2022-09-13
create backups when new schedule is created

birsanv:

Changes:
- when new schedules are created, generate new backups; this is the default behavior since we just moved to install oadp 1.1 in 2.7; which no longer creates backups on schedule creation
- you can disable the backup creation by setting `Spec.NoBackupOnStart=true` on the BackupSchedule CR


sahare (approved):
lgtm

PR #311, opened 2022-09-23
validate MSA option

birsanv:
https://github.com/stolostron/backlog/issues/24155

Show error if MSA is not enabled on the ACM MCH and the backup schedule enables the function

Screen Shot 2022-09-23 at 2 25 23 PM


sahare (approved):
Looks good, Thanks!

PR #313, opened 2022-09-26
use orSelector to create credentials backup

birsanv:

Changes:
Using the oadp 1.1 OrLabelSelectors []*metav1.LabelSelector , regrouped the 3 credential schedules under one schedule

With this change, we generate only one credentials backup now, although restore, for backup compatibility,  should be able to restore one or 3 backups format for the credentials
Screen Shot 2022-09-26 at 1 30 56 PM
Screen Shot 2022-09-26 at 1 31 17 PM


sahare (approved):
Looks good, Thanks!

PR #314, opened 2022-09-27
restore uses old hive cred backup whith oadp 1.1 orSelector

birsanv:

related to https://github.com/stolostron/cluster-backup-operator/pull/313

If the backup doesn't create the hive and cluster creds backup ( with oadp 1.1 we group them in one backup ), restore looks for hive and cluster creds backup and finds ones that should not match the current backup

Change : look for those backups, for backward compatibility, but get the backups that matches the `acm-credentials-schedule` by start time and timestamp.


sahare (approved):
lgtm

PR #320, opened 2022-09-30
refactor post restore code

birsanv:

- Move post restore code to a separate file in preparation for the new delta restore implementation
- setRestorePhase returns now a bool cleanupOnRestore which is true if the status of the restore has changed to Enabled - after a new restore was called on sync
- if cleanupOnRestore = true or (restoreStatus = RestorePhaseFinished || restoreStatus = RestorePhaseFinishedWithErrors) then run delta cleanup


sahare (approved):
lgtm, thanks!

PR #323, opened 2022-10-04
process delta for restore clusters

birsanv:

Clean up managed cluster backup resources post restore; remove resources created by a previous restore which are no longer part of the latest restored backup
If managedClusters are not restored, clean up generic resources but with a label value not equal with cluster-activation
If managedClusters are restored, clean up all generic resources


sahare (approved):
lgtm

PR #324, opened 2022-10-06
dont run clean up if option is set to CleanupTypeNone

birsanv:

https://github.com/stolostron/backlog/issues/26192

Changes:
- removed doc around `CleanupAll` option, it should not be used
- updated readme to refer to post restore instead of pre restore
- do not run cleanup if option is `CleanupTypeNone`


sahare (approved):
Looks good, Thanks!

PR #327, opened 2022-10-07
MSA component is now owned by MCE

birsanv:
https://github.com/stolostron/backlog/issues/23680

Update readme

MSA component is now owned by MCE


sahare (approved):
lgtm

PR #334, opened 2022-10-14
fix CRD not picked up after pod is running

birsanv:

https://github.com/stolostron/backlog/issues/24206#issuecomment-1279479335

Changes:
- reconcile after backupschedule failed validation due to the MSA option being invalid
- recreate the mapper on each reconcile
- remove mapper for restore reconciler, create one from discoveryInterface


sahare (approved):
Looks good, Thanks!

PR #340, opened 2022-10-18
don't backup MSA and clean up secret on new backup

birsanv:

https://github.com/stolostron/backlog/issues/26385

Don't backup MSA and manifest work
Before the MSA is created, delete any secrets with the same name ( if this is a passive - becoming active, there could be secrets created by the MSA on the initial primary; those secrets should be removed since this hub should enable MSA support before using the MSA feature; and creating new secrets with the same MSA won't work)


sahare (approved):
lgtm

PR #346, opened 2022-10-20
Set MSA as Tech Preview 

birsanv:
https://github.com/stolostron/backlog/issues/23583

Changes:
- Readme mark MSA as Tech Preview 
- add open-cluster-management-backup ns to the samples and readme; the chart always installs under this ns, regardless of the MCH ns


sahare (approved):
lgtm

PR #364, opened 2022-11-18
backup MSA and MSA addon

birsanv:

https://issues.redhat.com/browse/ACM-2074

Don't exclude MSA and MSA addon from backup


sahare (approved):
lgtm

PR #370, opened 2022-11-23
set keeping-auto-import-secret annotation

birsanv:

https://github.com/stolostron/backlog/issues/26911

related to fix https://github.com/stolostron/managedcluster-import-controller/pull/201

Set the keeping-auto-import-secret annotation on auto-import-secret so that the import controller retries the import if the cluster is not available at the time of the restore

Changes:
- remove readme limitation `If the token is valid when the restore operation is executed but the managed cluster is not accessible at the time the restore is executed, the auto import operation fails and will not retry to reconnect. In this case the failure should be reported by the auto-import component logs.`
- add the `managedcluster-import-controller.open-cluster-management.io/keeping-auto-import-secret` to the `auto-import-secret` created by the post restore

fyi @thuyn-581


tesshuflower (approved):
/lgtm

PR #376, opened 2022-12-12
don't backup hive import secrets

birsanv:

https://issues.redhat.com/browse/ACM-2327

Reported issue
Steps to Reproduce:

1. On prim hub, create clusterpool + claim a managed cluster >> OK
2. Hibernate the managed cluster claimed in step 1
3. Create backup schedule >> backups created successfully
4. Shutdown prim hub, run restore activation on second hub
5. Managed cluster restored on second hub in hibernating state
6. Resume the hibernating managed cluster in step 5

Actual results:
The resumed managed cluster is stuck at 'Pending import'

When we backup a cluster, we should not backup its import secret, we should let import controller create it in the new hub

### DIFF
--- a/controllers/pre_backup.go
+++ b/controllers/pre_backup.go
@@ -728,8 +728,27 @@ func updateSecretsLabels(ctx context.Context,
 	labelName string,
 	labelValue string,
 ) {
+	logger := log.FromContext(ctx)
+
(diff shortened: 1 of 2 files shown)


tesshuflower (approved):
/lgtm

PR #377, opened 2022-12-13
add restore spec excludedNamespaces and excludedResources

birsanv:

https://github.com/stolostron/backlog/issues/26919

Changes:
- Allow setting excludedNamespaces and excludedResources on the acm restore resource
- Pass this to the velero restore resources created by the acm restore resource
- update velero CRD with version 1.9.4 mapping to oadp 1.1.1
- I am setting the Spec.IncludeClusterResources=true on velero restore for all restores except the credentials; I noticed that with the excludedNamespaces option set on the restore backup, if the Spec.IncludeClusterResources is missing it is not beign defaulted to true as it is documented. I am setting it to true so that global resources are restored


tesshuflower (approved):
/lgtm

tesshuflower (approved):
/lgtm

PR #378, opened 2022-12-14
always attempt to create MSA manifest work

birsanv:

https://github.com/stolostron/backlog/issues/27027

Issue
Sometimes the MSA rolebinding is not lost on the managed cluster, even if the role still exists
Fix : attempt to  create the MSA manifest work on the hub even if the MSA addon exists


tesshuflower (approved):
/lgtm

PR #387, opened 2023-03-07
add optional volumeSnapshotLocation

birsanv:
https://issues.redhat.com/browse/ACM-4110

If the DPA resource defines more than one volume snapshot locations, the acm backup fails with this error

``` 
status:
expiration: '2023-03-12T14:02:11Z'
formatVersion: 1.1.0
phase: FailedValidation
validationErrors:

>-
provider aws has more than one possible volume snapshot location, and none
were specified explicitly or as a default
version: 1
 ```

Need to add an optional property to the BackupSchedule to set the volumeSnapshotLocation on the schedules, if more than one are defined

Even if the backup doesn't get any volumes, if more than one volumeSnapshotLocation is defined with the DPA, the backup throws an error if it's doesn't specify which volumeSnapshotLocation may be used

The PR adds this volumeSnapshotLocation optional property to the BackupSchedule so that it can be passed down to the velero schedules

Changes:

BackupSchedule has an optional property now used to pass the volumeSnapshotLocation

```
kind: BackupSchedule
metadata:
  name: example
  namespace: open-cluster-management-backup
spec:
  veleroSchedule: 0 */2 * * *
  veleroTtl: 120h
  volumeSnapshotLocations:
    - dpa-1
```

example of a DPA spec with 2 volume snapshot locations

```
spec:
  backupLocations:
    - velero:
        config:
          profile: default
          region: us-east-1
        credential:
          key: cloud
          name: cloud-credentials
        default: true
        objectStorage:
          bucket: velero-backup-acm-vb
          prefix: obs-0302-1
        provider: aws
  configuration:
    restic:
      enable: true
    velero:
      defaultPlugins:
        - openshift
        - aws
      podConfig:
        resourceAllocations:
          limits:
            cpu: '2'
            memory: 1Gi
          requests:
            cpu: 500m
            memory: 256Mi
  snapshotLocations:
    - velero:
        config:
          profile: default
          region: us-east-2
        provider: aws
    - velero:
        config:
          profile: east1
          region: us-east-1
        provider: aws
```


tesshuflower (approved):
/lgtm

PR #388, opened 2023-03-13
allow activation resources by group api

birsanv:
Allow specifying the apiGroup for resources that need to be restored at activation time

backup agent installer as activation data resource 

Changes:
- introduced a new group includedActivationAPIGroupsByName , similar with includedAPIGroupsByName
- resources with the api groups which are part of the includedActivationAPIGroupsByName list will be stored under the acm-managed-clusters-schedule and will be restored at activation time, with the managed cluster resources
- moved the agent-install.openshift.io under this group


tesshuflower (approved):
/lgtm

PR #390, opened 2023-03-16
use go install for getting build dependencies

tesshuflower:
- also set HOME only in build target inside Makefile.prow so it doesn't override for everything


birsanv (approved):
lgtm

birsanv (approved):
lgtm

PR #400, opened 2023-04-12
Update leader election flag names - use duration

tesshuflower:
For: https://issues.redhat.com/browse/ACM-4767

Updating flags from previous PR - using standard examples from:  https://issues.redhat.com/browse/ACM-4108

Changed: flags are now using the standard names from the above examples and are now durations.


sahare (approved):
Looks good, Thanks!

PR #402, opened 2023-06-15
use generic backup name on cleanup

birsanv:
https://issues.redhat.com/browse/ACM-5953

New output for restore (veleroGenericResourcesRestoreName); this will be used to find the backup name for the generic resources. Before this change, the generic resources backup name was computed from the resources backup and this could result in cleaning up generic resources  if the resources backup was not from the same schedule with the generic backup ( in that case, newly restored generic resources were identified as restored in another session and were cleaned by the post restore operation )

```
kind: Restore
metadata:
  name: restore-acm-passive-sync
  namespace: open-cluster-management-backup
spec:
  cleanupBeforeRestore: CleanupRestored
  restoreSyncInterval: 5m
  syncRestoreWithNewBackups: true
  veleroCredentialsBackupName: latest
  veleroManagedClustersBackupName: skip
  veleroResourcesBackupName: latest
status:
  lastMessage: >-
    Velero restores have run to completion, restore will continue to sync with
    new backups
  phase: Enabled
  veleroCredentialsRestoreName: restore-acm-passive-sync-acm-credentials-schedule-20230615200051
  veleroGenericResourcesRestoreName: restore-acm-passive-sync-acm-resources-generic-schedule-20230615200051
  veleroResourcesRestoreName: restore-acm-passive-sync-acm-resources-schedule-20230615200051
```

### DIFF
--- a/api/v1beta1/restore_types.go
+++ b/api/v1beta1/restore_types.go
@@ -137,6 +137,8 @@ type RestoreStatus struct {
 	// +kubebuilder:validation:Optional
 	VeleroResourcesRestoreName string `json:"veleroResourcesRestoreName,omitempty"`
 	// +kubebuilder:validation:Optional
+	VeleroGenericResourcesRestoreName string `json:"veleroGenericResourcesRestoreName,omitempty"`
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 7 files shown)


sahare (approved):
lgtm

PR #413, opened 2023-08-02
backup multiclusterobservability under active data backup

birsanv:
https://issues.redhat.com/browse/ACM-4225

Move MultiClusterObservability to the active resource set since this resource once restored on the restore hub creates all obs resources using the backed secrets and configmaps; as a result it creates PV and PVClaims
We don't want to have these resources created until the very end, when the passive cluster becomes active


sahare (approved):
lgtm

PR #421, opened 2023-08-16
remove MSA installNS on ManagedClusterAddOn

tesshuflower:
Do not set spec.installNamespace on managed-serviceaccount ManagedClusterAddOn when creating it.

For: https://issues.redhat.com/browse/ACM-6967

### DIFF
--- a/controllers/pre_backup.go
+++ b/controllers/pre_backup.go
@@ -247,9 +247,7 @@ func prepareImportedClusters(ctx context.Context,
 			}
 
 			alreadyCreated := false
-			installNamespace := ""
 			for addon := range addons.Items {
-				installNamespace = addons.Items[addon].Spec.InstallNamespace
 				if addons.Items[addon].Name == msa_addon {
 					alreadyCreated = true
 					break
@@ -259,7 +257,7 @@ func prepareImportedClusters(ctx context.Context,
 				msaAddon := &addonv1alpha1.ManagedClusterAddOn{}
 				msaAddon.Name = msa_addon
 				msaAddon.Namespace = managedCluster.Name
-				msaAddon.Spec.InstallNamespace = installNamespace
+				// Not setting msaAddon.Spec.InstallNamespace - will leave default
 				labels := map[string]string{
 					msa_label: msa_service_name,
 				}
--- a/controllers/schedule_controller_test.go
+++ b/controllers/schedule_controller_test.go
@@ -460,6 +460,37 @@ var _ = Describe("BackupSchedule controller", func() {
 					autoImportSecret.GetLabels()["cluster.open-cluster-management.io/backup"] == "msa"
 			}, timeout, interval).Should(Equal(rhacmBackupSchedule.Spec.UseManagedServiceAccount))
 
+			// validate that the managedserviceaccount ManagedClusterAddOn is created for managed clusters since
+			// useManagedServiceAccount is true
+			var managedSvcAccountMCAOs []addonv1alpha1.ManagedClusterAddOn
+			for i := range managedClusters {
+				// managed-serviceaccount ManagedClusterAddOn should not be created for local-cluster
+				if managedClusters[i].GetName() != "local-cluster" {
+					managedSvcAccountMCAOs = append(managedSvcAccountMCAOs, addonv1alpha1.ManagedClusterAddOn{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      msa_addon,
+							Namespace: managedClusters[i].GetName(),
+						},
+					})
+				}
+			}
+			Eventually(func() error {
+				for i := range managedSvcAccountMCAOs {
+					err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&managedSvcAccountMCAOs[i]),
+						&managedSvcAccountMCAOs[i])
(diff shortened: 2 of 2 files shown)


tesshuflower:
@zhujian7  Could I get you to take a look at the change in pre_backup.go and let me know if this is ok?  Essentially if the managedclusteraddon for managed-serviceaccount does not already exist, this change will create a new one and NOT set spec.installNamespace.

This assumes the managed-serviceaccount ManagedClusterAddOn controller will handle this and properly set the default. Is that assumption correct?

controllers/pre_backup.go
```diff
@@ -259,7 +257,7 @@ func prepareImportedClusters(ctx context.Context,
 				msaAddon := &addonv1alpha1.ManagedClusterAddOn{}
 				msaAddon.Name = msa_addon
 				msaAddon.Namespace = managedCluster.Name
-				msaAddon.Spec.InstallNamespace = installNamespace
+				// Not setting msaAddon.Spec.InstallNamespace - will leave default
```
zhujian7, line 260: I think keep it empty is ok since there is a [default value](https://github.com/open-cluster-management-io/api/blob/2b2399b5f6e88e28a962aa82307ccc00bc492009/addon/v1alpha1/types_managedclusteraddon.go#L39): "open-cluster-management-agent-addon" (added from ACM 2.7)

controllers/pre_backup.go
```diff
@@ -259,7 +257,7 @@ func prepareImportedClusters(ctx context.Context,
 				msaAddon := &addonv1alpha1.ManagedClusterAddOn{}
```
zhujian7, line 257: Is it possible to also update the install namespace of the existing managed-serviceaccount addon created by the backup operator?

zhujian7: I saw there is a label in backup-operator code, line 261 - line 263:
```
				labels := map[string]string{
					msa_label: msa_service_name,
				}
```
If this can indicate that the managed-serviceaccount addon was created by DR?

tesshuflower: @zhujian7 the problem is that this label is kind of generic (it's just "authentication.open-cluster-management.io/is-managed-serviceaccount": "auto-import-account") so I'm not sure if it can be added by others as well.  I don't know this history of why this label was applied.

There is also still the question of whether it's safe for us to change the namespace.  Is this something that won't cause any issues?  If the namespace must be fixed, I'd prefer this is implemented & tested in the managed service account addon itself, rather than backup&restore needing to figure out all the permutations of changing on clusters  out there.  I'm not clear on whether we can simply change the installNamespace on this addon (I assume we'd reset it to "") if it's already set to something non-standard.

zhujian7: Hi @tesshuflower , sorry to make you confused, let me give some background and try to answer your questions:

> There is also still the question of whether it's safe for us to change the namespace. Is this something that won't cause any issues?

I tested changing the install namespace, it will delete the agent in the old ns, and create another one in the new ns. and **the old ns (in the customer case "openshift-storage") will be deleted**. but after changing it to "open-cluster-management-agent-addon", deleting the addon will not trigger to delete the "open-cluster-management-agent-addon" ns, once [this](https://github.com/open-cluster-management-io/managed-serviceaccount/pull/74) is merged. (I can test again tomorrow if you're worried)

> If the namespace must be fixed, I'd prefer this is implemented & tested in the managed service account addon itself

For 2.7 and 2.8, it is not required to be a fixed value, it is configurable, but deleting the addon will trigger to delete the install ns, so I submit a PR to [not delete the ns if it is the default ns](https://github.com/stolostron/managed-serviceaccount/pull/36) to address the customer issue we have now. why do not we orphan any ns no matter what name it is but just the "open-cluster-management-agent-addon", because the namespace created by enabling the managed-serviceaccount addon should be deleted when disabling the addon, for example, if the install ns is set to "open-cluster-management-service-account", enabling the addon will create the ns on the managed cluster, it should not remain when deleting the addon. So we recommend changing the ns to empty/default "open-cluster-management-agent-addon".


For 2.9, because we are planning to change it to a template-type addon, unfortunately, there are some issues we can not address now, as a result, we want to only support a fixed value in 2.9, but in the future, it should be configurable. And I think it is still not decided to change it to a fixed value before Train 6 starts, I am collecting feedback from managed serviceaccount users, if you strongly disagree, I think we can stop changing it. cc @qiujian16 

> rather than backup&restore needing to figure out all the permutations of changing on clusters out there.

sorry, I don't quite understand this, Could you give more explanations?

I hope this helps. And any suggestions or other questions are welcome.

zhujian7:
> I think we were hoping to commit this change relatively soon as the ODF team is doing testing using pre-release ACM 2.9 and would like the change. 

OK, got it, I am working on it [for 2.9](https://github.com/open-cluster-management-io/managed-serviceaccount/pull/74) train 5

zhujian7:
> I am working on it https://github.com/open-cluster-management-io/managed-serviceaccount/pull/74 train 5

based on the new [discussion](https://issues.redhat.com/browse/ACM-6864?focusedId=22838608&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22838608), I don't think we can fix this in Train 5

PR #432, opened 2023-09-13
MSA CR is set to beta1

birsanv:
https://issues.redhat.com/browse/ACM-7481

dynamic go client throws an exception when trying to create an MSA as v1alpha1
Changed the code to use the new v1beta1


sahare (approved):
lgtm

sahare (approved):
lgtm

PR #443, opened 2023-10-25
create role binding for custom and default ns

birsanv:
https://issues.redhat.com/browse/ACM-8348

https://issues.redhat.com/browse/ACM-8263

Issue:
1. in 2.8.2 For MSA accounts created in a ns different than the default addon, we create a custom ManifestWork to create a RoleBinding for the ServiceAccount in that ns

The issue is that the RoleBinding has the same name as the one created by the initial ManifestWork and since this is a global resource, both ManifestWork owns it. Each ManifestWork keep updating the RoleBinding 

 
```
oc get clusterrolebinding managedserviceaccount-import -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
creationTimestamp: "2023-10-25T15:42:54Z"
name: managedserviceaccount-import
ownerReferences:

apiVersion: work.open-cluster-management.io/v1
kind: AppliedManifestWork
name: 13583c9e64c6bd8e94c94f7aa0e5a3b873c8033fc07c1f2ceb7c6b5da3056750-addon-managed-serviceaccount-import
uid: b5fcf244-e461-404a-88c7-4c4aeba5db57
apiVersion: work.open-cluster-management.io/v1
kind: AppliedManifestWork
name: 13583c9e64c6bd8e94c94f7aa0e5a3b873c8033fc07c1f2ceb7c6b5da3056750-addon-managed-serviceaccount-import-custom
uid: c0476c37-c3b0-4184-bedd-e9ffb10e131c
resourceVersion: "49530"
uid: 597859e7-ce78-43dc-8b65-b038492751bc
roleRef:
apiGroup: rbac.authorization.k8s.io
kind: ClusterRole
name: klusterlet-bootstrap-kubeconfig
subjects:
kind: ServiceAccount
name: auto-import-account
namespace: open-cluster-management-addon-observability
```

2. second issue for 2.9 : even when a MSA is to use a custom ns, the framework ignores this value and puts in into the default addon ns. The backup operator code is not aware of this change and it creates the custom ManifestWork, which creates the RoleBinding for the custom ns, instead of the default one 

Fixes:
- Delete the custom Manifest work if exists
- if the MSA sets a custom ns, create a custom-2 suffixed ManifestWork ; use the same name for the RoleBinding in the workload so it doesn't collide with the one created for the addon ns
- Even if the MSA is set to a custom ns, to account for the framework fixes in 2.9, we also create a ManifestWork for the default addon ns

### DIFF
--- a/controllers/create_helper.go
+++ b/controllers/create_helper.go
@@ -8,6 +8,7 @@ import (
 	v1beta1 "github.com/stolostron/cluster-backup-operator/api/v1beta1"
 	veleroapi "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	workv1 "open-cluster-management.io/api/work/v1"
 	chnv1 "open-cluster-management.io/multicloud-operators-channel/pkg/apis/apps/v1"
 )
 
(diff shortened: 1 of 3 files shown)


tesshuflower (approved):
/lgtm

PR #451, opened 2023-10-31
remove obs secret on post restore

birsanv:
https://issues.redhat.com/browse/ACM-8406

Remove this secret as part of the post restore operation
`oc get secret -n open-cluster-management-addon-observability        observability-controller-open-cluster-management.io-observability-signer-client-cert`


tesshuflower (approved):
/lgtm

PR #460, opened 2023-12-06
exclude CRDs from restore

birsanv:
https://issues.redhat.com/browse/ACM-8904

https://issues.redhat.com/browse/ACM-8870

Exclude CRDs when restoring a backup using the `Restore.Spec.ExcludedResources` option


tesshuflower (approved):
/lgtm

PR #468, opened 2024-01-18
active configmaps not restored with no sync option

birsanv:
related to https://github.com/stolostron/cluster-backup-operator/pull/467

Fixed this issue with ConfigMaps set to be restored on cluster-activation:

The resources were not restored if the veleroCredentialsBackupName: skip ( so using the no-sync approach )https://github.com/stolostron/cluster-backup-operator/blob/main/config/samples/cluster_v1beta1_restore_passive.yaml#L13


sahare (approved):
lgtm

PR #469, opened 2024-01-18
Exclude namespace resources from generic backups as a temporary workaround

sahare:
Jira https://issues.redhat.com/browse/ACM-9457 

Until a fix for the following OADP issue is available, as a workaround we exclude namespace resources from generic backups by adding them under excludedResources. This problem is introduced in OADP 1.3 and will be fixed in OADP future releases so we will remove the workaround as soon as we have an OADP fix.

OADP issue https://issues.redhat.com/browse/OADP-3364


birsanv:
@sahare can you fix the code issue first

sahare:
@birsanv please review, thanks!

birsanv (approved):
lgtm thanks!

PR #481, opened 2024-01-30
wait for pvc when only creds files were restored

birsanv:
https://issues.redhat.com/browse/ACM-9717

If a configMap with 
cluster.open-cluster-management.io/backup-pvc label is found during a restore operation, the backup operator must wait for the PVC to be created ( by the pvc policy , which also restores the snapshot using a volsync DestinationReplication) , before creating activation resources. This allow the PVC to be restored before the app using the pvc uses the volume.
The issue reported here : The operator waits for the PVC but if, during the wait, another credentials backup is created by the primary hub , the new credential is picked up and because the wait stops.
The wait should not resume if only credential backups have been restored.

Version-Release number of selected component (if applicable):
How reproducible:
Always

Steps to Reproduce:

1. On the primary hub ( have the pv policy installed ) then set the  cluster.open-cluster-management.io/backup-hub-pvc label on a PVC so the configmap with  cluster.open-cluster-management.io/backup-pvc is created and backed up
2.  Run an acm backup schedule on the restore hub to make sure the data is backed up
3.  Go to the restore hub:

- On the restore hub remove the pvc policies so that after a restore you get into the wait stage
- Run a restore all operation
- You should see the Restore status saying : wait for PVC to be created.. ; and shows the. credentials file restored 
- Now go back on the backup hub and trigger a new backup ( delete one of the velero schedules , this should trigger that )

4. Go back to the restore hub and wait until the new backup is sync up and picked up by the restore op ( which uses latest as the option to restore )
5. You should see at some point in the acm restore status that the new credentials backup was restored ( you can also check that you see the velero restore resource )
6.  Notice the wait is resumed and all resources are restored, even though the PVC was not created yet
7. Notice the wait is resumed and all resources are restored, even though the PVC was not created yet 

Actual results:
Notice the wait is resumed and all resources are restored, even though the PVC was not created yet 

Expected results:
Step 7 should not happen; the restore should still say wait until the PVC is created, even if a second credentials backup was restored.

### DIFF
--- a/controllers/create_helper.go
+++ b/controllers/create_helper.go
@@ -268,6 +268,11 @@ func (b *RestoreHelper) backupName(name string) *RestoreHelper {
 	return b
 }
 
+func (b *RestoreHelper) scheduleName(name string) *RestoreHelper {
+	b.object.Spec.ScheduleName = name
+	return b
+}
+
 func (b *RestoreHelper) phase(phase veleroapi.RestorePhase) *RestoreHelper {
 	b.object.Status.Phase = phase
 	return b
--- a/controllers/create_helper.go
+++ b/controllers/create_helper.go
@@ -2,6 +2,8 @@ package controllers
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


sahare (approved):
lgtm

sahare (approved):
lgtm

PR #489, opened 2024-02-20
run cleanup step when waitPVC is false

birsanv:
https://issues.redhat.com/browse/ACM-9891

Related to https://issues.redhat.com/browse/ACM-9717
The fix : run the post restore steps once the restore is not waiting on the PVC creation


sahare (approved):
Looks good, Thanks!

PR #490, opened 2024-02-22
restore msa ManifestWork at cluster activation

birsanv:
related to https://issues.redhat.com/browse/ACM-9780

Any backed up ManifestWork should be restored at activation time, after managed cluster CRs


sahare (approved):
lgtm

PR #494, opened 2024-02-27
filter out MSA tokens not used by auto-import

birsanv:
https://issues.redhat.com/browse/ACM-10093

Issue: When restoring managed clusters data and auto-import-account MSA are used to connect to the clusters, the list of MSA used by the restore operation should only include tokens created by an MSA with the auto-import-account name
If not, other MSA secrets can be picked up to be used by the auto-import-secret and those tokens don't have authority to run the import operation

Fix:
When looking for MSA tokens filter out and get only the ones with a name prefix `auto-import-account`

### DIFF
--- a/controllers/pre_backup.go
+++ b/controllers/pre_backup.go
@@ -626,7 +626,7 @@ func getMSASecrets(
 			if err := c.List(ctx, msaSecrets, &client.ListOptions{
 				LabelSelector: selector,
 			}); err == nil {
-				return msaSecrets.Items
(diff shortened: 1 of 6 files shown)


xuezhaojun:
Hi, @birsanv I create a sperate Jira to trace the permission error which only affect MCE 2.5.0, and left the orign Jira to trace the auth error which affect MCE 2.5.0 and before releases.

Would mind rediect the PR to the new Jira https://issues.redhat.com/browse/ACM-10156 thanks!

sahare (approved):
lgtm

PR #509, opened 2024-04-02
Remove managedcluster.clusterview.open-cluster-management.io from Backup

sahare:
Jira: https://issues.redhat.com/browse/ACM-10638


birsanv (approved):
lgtm

We also need to update the docs if this is documented there

PR #510, opened 2024-04-04
Ubi9 and konflux dockerfile

tesshuflower:
Update upstream dockerfile to use ubi9 and also checkin Dockerfile.rhtap for konflux.

Committing to main and will ffwd to release-2.11


sahare (approved):
lgtm

PR #534, opened 2024-05-21
dependent crd checks at startup

tesshuflower:
If Velero CRDs are missing, starting controllers will crash as they have a watch on velero Schedule and Restore resources.  Adding a check that only happens prior to controller startup so we can wait for the CRDs to appear.

In a loop, checking every 30s at startup (before starting manager & controllers):
- Check for Velero Schedule and Restore CRDs at startup
- If either are not present, do not start controllers


birsanv (approved):
lgtm

PR #536, opened 2024-05-22
move dependent crds to hack/crds dir

tesshuflower:
Moving crds that are not generated by cluster-backup and only needed for testing so it's obvious which crds are generated by us (and make targets that work on them should work more as intended as well)


birsanv (approved):
lgtm

PR #538, opened 2024-05-27
show backup collision if restore activation run on another hub

birsanv:
https://issues.redhat.com/browse/ACM-11876

related to https://github.com/stolostron/cluster-backup-operator/pull/537

Scenario:

- hubA is active, backup schedules running
- hubB is passive
- hubA goes down unexpectedly 
- admin restores managed clusters on hubB and hubB becomes active hub
- hubA starts up; the backup schedule is still enabled
- since hubB is now the active hub, hubA schedule should be disabled

Changes : 
If a BackupSchedule runs on this hub, on reconcile, check if any acm-restore-clusters backup resource was created after the BackupSchedule was created. 
If a acm-restore-clusters backup resource was found, created after the BackupSchedule was created, and created by a different hub, set the BackupSchedule to BackupCollision and remove velero schedules

### DIFF
--- a/controllers/schedule.go
+++ b/controllers/schedule.go
@@ -48,11 +48,11 @@ const (
 		" This is a backup collision with current cluster [%s] backup." +
 		" Review and resolve the collision then create a new BackupSchedule resource to " +
 		" resume backups from this cluster."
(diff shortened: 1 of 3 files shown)


tesshuflower (commented):
/lgtm

This seems like a really nice feature!

PR #551, opened 2024-06-03
Some unit test updates

tesshuflower:
- to test restore reconcile - previously there could be a timing issue (see PR:
    https://github.com/stolostron/cluster-backup-operator/pull/547)

- Before the above PR, this test would fail approximately 4 out of 5 attempts


birsanv (approved):
lgtm
thanks Tesshu!

PR #650, opened 2024-07-31
add support for velero restore filter options

birsanv:
https://issues.redhat.com/browse/ACM-12973

Enable ACM restore support for the following velero restore options:
see https://velero.io/docs/main/api-types/restore/ CRD
- includedResources
- includedNamespaces
- OrLabelSelectors
- LabelSelector
- namespaceMapping

Added a sample that shows how to use some of the velero restore options when restoring the  ACM hub backup
We strongly recommend to review the cleanupBeforeRestore option and use cleanupBeforeRestore=None when restoring a subset of the data, as using cleanupBeforeRestore=CleanupRestored may clean up resources excluded by the restore filter criteria.

### DIFF
--- a/controllers/restore.go
+++ b/controllers/restore.go
@@ -779,6 +779,10 @@ func setOptionalProperties(
 			veleroRestore.Spec.LabelSelector = labels
 		}
 
+		// append LabelSelector resources since the acm restore uses the veleroRestore.Spec.LabelSelector
+		// to filter out activation resources when restoring the data - as it does for credentials-active restore, using the cluster.open-cluster-management.io/backup=cluster-activation label
+		// we want to keep any acm predefined veleroRestore.Spec.LabelSelector and add to those the user defined ones
+
 		// set MatchExpressions
 		if acmRestore.Spec.LabelSelector.MatchExpressions != nil {
 			// create the MatchExpression for the velero resource, if not defined yet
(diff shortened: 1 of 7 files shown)


controllers/restore.go
```diff
@@ -768,6 +770,45 @@ func setOptionalProperties(
 			acmRestore.Spec.Hooks.Resources...,
 		)
 	}
+
+	// add any label selector set using the acm restore resource spec
+	if acmRestore.Spec.LabelSelector != nil {
+
+		if veleroRestore.Spec.LabelSelector == nil {
```
tesshuflower, line 777: Just for my understanding: Is there a reason we don't set veleroRestore.Spec.LabelSelector to acmRestore.Spec.LabelSelector?  Is velero or oadp appending their own labelselectors that we then have to modify?

birsanv: acmRestore resource creates the velero restores using the options set on the acmRestore; the velero restore options above were not surfaced to the user so there was no way for a user creating and acm restore resource to set these options for the velero restores

tesshuflower: Just wondering why we don't do:

```
veleroRestore.Spec.LabelSelector = acmRestore.Spec.LabelSelector 
```

like you do later on with OrLabelSelectors as an example.  This way any existing label selectors will be kept, rather than reconciling to the exact labelSelectors specified in the acmRestore - but perhaps that is desired for some reason?

birsanv: We don't set the veleroRestore.Spec.LabelSelector to acmRestore.Spec.LabelSelector because the operator sets a 
veleroRestore.Spec.LabelSelector when creating the credentials restore file, to exclude resources that should be restored only when the activation step is executed ; we want to keep any predefined veleroRestore.Spec.LabelSelector values and add to those the ones defined by the user using the acmRestore spec

The reason I didn't do the same for the OrLabelSelectors is because we are not using this when creating the restore resource. 
But it makes sense to do the same, to be consistent and for the future, if we decide to use the OrLabelSelectors option, so good point.

I updated the code description to explain why I am appending to the veleroRestore.Spec.LabelSelector and changed the veleroRestore.Spec.OrLabelSelector to append as well.

birsanv: That's another good question : velero restore files once created, they go to completion and cannot be updated.
So the scenario could be : 
1. the user runs a one time acm restore ( syncRestoreWithNewBackups: false ) : this acm restore creates the velero restore files, using the options set by the user. Once this is done, acm restore cannot be modified, the status goes to Finished once all velero restores are completed. Velero restores cannot be updated once created.
2. the user runs a passive acm restore ( syncRestoreWithNewBackups: true, a new set of velero restore will run at the specified interval ). In this case, again Velero restores cannot be updated once created. The user can update the acm restore options ( add / remove labels, etc ) but this gets reflected on the new set of restore files, for the next run. 

I put the PR on hold to fix the issue where the function has too many lines, I'll change this next

tesshuflower: ah, maybe that makes it simpler in the end if we don't need to update existing velero restores - if we only do this at creation time it becomes simpler (this stuff is so hard to get right) as we just use the user settings at creation time.

PR #655, opened 2024-08-06
ignore orLabelSelector for credentials and generic resources restore

birsanv:
Related to https://github.com/stolostron/cluster-backup-operator/pull/650

https://issues.redhat.com/browse/ACM-12973

Changes:
When the orLabelSelector is used to filter out restored objects, the orLabelSelector will be ignored for the credentials and generic resources velero restore files, since they already use a labelSelector ( set by the backup operator ) to filter out resources with a cluster-activation label
LabelSelector and OrLabelSelectors are mutually exclusive so you can't use both

labelSelector set by ACM on the velero restore resource for credentials and generic resources:

```
  labelSelector:
    matchExpressions:
      - key: cluster.open-cluster-management.io/backup
        operator: In
        values:
          - cluster-activation
```

### DIFF
--- a/controllers/restore.go
+++ b/controllers/restore.go
@@ -773,7 +773,7 @@ func setOptionalProperties(
 	}
 
 	// set user options for resource filtering
-	setUserRestoreFilters(acmRestore, veleroRestore)
+	setUserRestoreFilters(acmRestore, veleroRestore, key)
 
 	// allow namespace mapping
 	if acmRestore.Spec.NamespaceMapping != nil {
@@ -785,6 +785,7 @@ func setOptionalProperties(
 func setUserRestoreFilters(
 	acmRestore *v1beta1.Restore,
 	veleroRestore *veleroapi.Restore,
+	key ResourceType,
 ) {
 
 	// add any label selector set using the acm restore resource spec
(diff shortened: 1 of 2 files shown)


controllers/restore.go
```diff
@@ -826,7 +827,9 @@ func setUserRestoreFilters(
 	}
 
 	// add any or label selector set using the acm restore resource spec
-	if acmRestore.Spec.OrLabelSelectors != nil {
+	// skip resources generic and credentials since they already define a LabelSelector for cluster activation
+	// LabelSelector and OrLabelSelectors are mutually exclusive
+	if (key != ResourcesGeneric && key != Credentials) && acmRestore.Spec.OrLabelSelectors != nil {
```
tesshuflower, line 832: Do you think it would be possible to do this in a more generic way so we don't need to hardcode for ResourcesGeneric and Credentials?

I think if I understand correctly by this point in the code, we've already set LabelSelectors - either ones that we, cluster-backup set like in the case of `ResourcesGeneric` or `Credentials`, or ones the user possibly sets by using their own LabelSelector in the ACMRestore spec.

Could we instead only set OrLabelSelectors if LabelSelectors is not set?  Would this cover all cases?

sahare (approved):
lgtm

controllers/restore.go
birsanv: Not really, Tesshu, but we can discuss off line

PR #665, opened 2024-08-12
allow pausing the backup schedule

birsanv:
https://issues.redhat.com/browse/ACM-13242

Changes: 
- introduces a new optional, BackupSchedule spec option, named `paused`; default value is false. If paused=true, the BackupSchedule doesn't generate any velero schedules, so is not generating any backups. When an existing, Enabled, BackupSchedule changes to puase=true, all generated velero Scedules are deleted and the status of the BackupSchedule is changed to Paused
- a BackupSchedule is considered not to be running if BackupSchedule is Paused
- you can execute restore operations on the hub if the BackupSchedule is Paused
- you can create a BackupSchedule with paused=true if a restore is still active
- when checking for a BackupCollision situation, instead of verifying if the BackupSchedule creationTimestamp is older than the last cluster restore operation, it checks if the velero.Schedule ( created by the BackupSchedule) creationTimestamp is older than the last cluster restore operation. This is because the BackupSchedule creationTimestamp is no longer relevant to know when a backup was initiated, since the BackupSchedule can be paused ( removing the velero schedules, so not generating backups ) and then unpaused ( starting generating velero schedules and backups )


api/v1beta1/schedule_types.go
```diff
@@ -87,6 +89,11 @@ type BackupScheduleSpec struct {
 	// +kubebuilder:validation:Optional
 	// VolumeSnapshotLocations is a list containing names of VolumeSnapshotLocations associated with this backup.
 	VolumeSnapshotLocations []string `json:"volumeSnapshotLocations,omitempty"`
+	// +kubebuilder:validation:Optional
+	// When set to false, all velero Schedules generated by this BackupSchedule are being removed.
```
tesshuflower, line 93: Should this be when set to `true` all velero schedules will be removed?

And then when paused = false (the default), we go through the normal mode of creating backup schedules?

controllers/utils.go
```diff
@@ -470,3 +471,46 @@ func addRestoreLabelSelector(
...
+
+	err := c.Status().Update(ctx, backupSchedule)
+
+	// delete schedules, so we don't generate new backups
+	for i := range veleroScheduleList.Items {
+		scheduleLogger.Info(
+			"Attempt to delete schedule " + veleroScheduleList.Items[i].Name,
+		)
+
+		if err := c.Delete(ctx, &veleroScheduleList.Items[i]); err == nil {
```
tesshuflower, line 508: I think we should check for err != nil here and return the error if it occurs - otherwise if there's an error deleting a veleroschedule we may not retry until the next reconcile (whenever that happens)

controllers/utils.go
```diff
@@ -470,3 +471,46 @@ func addRestoreLabelSelector(
...
+
+	//update status
+	backupSchedule.Status.Phase = phase
+	backupSchedule.Status.LastMessage = msg
+
+	backupSchedule.Status.VeleroScheduleCredentials = nil
+	backupSchedule.Status.VeleroScheduleManagedClusters = nil
+	backupSchedule.Status.VeleroScheduleResources = nil
+
+	err := c.Status().Update(ctx, backupSchedule)
```
tesshuflower, line 500: Should we do the status update only after all the velero schedules have been successfully deleted? (just after the delete over veleroScheduleList.Items below) That way you know when you see the status is paused, we're done pausing.

api/v1beta1/schedule_types.go
birsanv: correct

controllers/utils.go
birsanv: I moved the BackupSchedule update after the schedules deletion; the code is being called only after the schedules are being successfully deleted.

controllers/utils.go
birsanv: I am checking if the schedule throws an error, and ignore the resource not found error. If any other delete errors, I call a reconcile and not update the BackupSchedule status

api/v1beta1/schedule_types.go
birsanv: Is there anything here that you suggest to be updated Tesshu ?
This is the current description 

```
	// +kubebuilder:validation:Optional
	// When set to false, all velero Schedules generated by this BackupSchedule are being removed.
	// Setting this option to true results in recreating the velero Schedules.
	// If not defined, the value is set to false.
	Paused bool `json:"paused,omitempty"`
```

controllers/utils.go
```diff
@@ -470,3 +472,59 @@ func addRestoreLabelSelector(
...
+			// ignore not found errors
+			if !kerrors.IsNotFound(err) {
+				errMsg := "Failed to delete schedule " + veleroScheduleList.Items[i].Name
+				scheduleLogger.Error(err,
+					errMsg,
+				)
+				// update failed, retry
+				backupSchedule.Status.LastMessage = errMsg
+				c.Status().Update(ctx, backupSchedule)
+				return ctrl.Result{RequeueAfter: failureInterval}, err
```
tesshuflower, line 514: I don't think you need to bother requeing after a specific failureInterval - the build in backoff should take care of this if you just return the error.  See: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

That way if something like the api server is unavailable, it'll take care of retrying less frequently (if there are multiple errors in a row) until hopefully things come back.  Essentially you can just do this:
```
return ctrl.Rresult{}, err
```

controllers/utils.go
```diff
@@ -470,3 +472,59 @@ func addRestoreLabelSelector(
...
+
+	//update status
+	backupSchedule.Status.Phase = phase
+	backupSchedule.Status.LastMessage = msg
+
+	backupSchedule.Status.VeleroScheduleCredentials = nil
+	backupSchedule.Status.VeleroScheduleManagedClusters = nil
+	backupSchedule.Status.VeleroScheduleResources = nil
+
+	c.Status().Update(ctx, backupSchedule)
```
tesshuflower, line 527:
```suggestion
	err := c.Status().Update(ctx, backupSchedule)
	return ctrl.Result{}, err
```

Suggest we check for errors in case the status update fails, that way if we didn't manage to update the status to "paused", it'll retry and we'll get there.

api/v1beta1/schedule_types.go
tesshuflower: I think you just have "true" and "false" switched in the comment.

when paused = "true", then we should be removing velero schedules and when paused = "false" (the default), then it should recreate the velero schedules.
```suggestion
	// When set to true, all velero Schedules generated by this BackupSchedule will be removed.
	// Setting this option to false results in recreating the velero Schedules.
	// If not defined, the value is set to false.
```

tesshuflower (approved):
/lgtm

PR #669, opened 2024-08-15
make a copy of user labels

birsanv:
https://issues.redhat.com/browse/ACM-13365

Set user label selectors on the velero restores by using a copy of the acm restore labels.
If using the acm restore labels directly, any update to this label would be reflected on all resources sharing the label selector - acm and velero restores.


tesshuflower (approved):
/lgtm

PR #679, opened 2024-08-21
Acm 13283 local cluster label

tesshuflower:
- updates to have helper functions to look for local cluster by looking at the managedcluster resources and finding one with the "local-cluster": "true" label/value.
- Includes updates to error handling
- includes unit test updates


birsanv (approved):
lgtm

PR #689, opened 2024-08-28
update to controller-gen v0.16.1 (https://issues.redhat.com/browse/ACM-13685)

tesshuflower:
https://issues.redhat.com/browse/ACM-13685

Old version of controller-gen (from controller-tools) was at v0.4.1 which doesn't compile with golang 1.22 - so not possible to run the make manifests or make generate build targets anymore.

Notes:
- Updated to the latest controller-gen (controller-tools v0.16.1)
- When running make manifests, some old CRD generate options are now gone.  We previously used the following:
```
controller-gen crd:preserveUnknownFields=false,crdVersions=v1,trivialVersions=true rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```

But now the following crd generate flags aren't supported anymore: `preserveUnknownFields` and `trivialVersions`.  Note also `crdVersions=v1` appears to be the default.  

So in the change I removed the options entirely so we could still run make manifests - note this does alter the crds slightly.


birsanv (approved):
lgtm

birsanv:
I'll open a PR to update the CRDs in the MCH project

PR #697, opened 2024-08-29
err handling when dynamic resource is not listable - For: https://issues.redhat.com/browse/ACM-13692

tesshuflower:
- we just ignore these, but need not to fail

Fixes: https://issues.redhat.com/browse/ACM-13692


birsanv (approved):
lgtm

PR #698, opened 2024-08-30
check velero restore PartiallyFailed (https://issues.redhat.com/browse/ACM-13693)

birsanv:
https://issues.redhat.com/browse/ACM-13693

Issue 
The velero restore status could return a FinalizingPartiallyFailed status for a restore that would eventually get into a PartiallyFailed status
The code only checks for the PartiallyFailed status so when the velero restore was FinalizingPartiallyFailed, acm restore returned success

Fix
Updated the code to check if the velero status contains the FailedStatus string

### DIFF
--- a/controllers/restore.go
+++ b/controllers/restore.go
@@ -234,7 +234,9 @@ func setRestorePhase(
 			)
 			return restore.Status.Phase, cleanupOnEnabled
 		}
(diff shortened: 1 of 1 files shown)


tesshuflower (approved):
/lgtm

PR #705, opened 2024-09-03
ACM-13001 - Set CompletionTimestamp when ACM restore is completed

sahare:
https://issues.redhat.com/browse/ACM-13001


birsanv (approved):
lgtm

PR #736, opened 2024-09-26
mark step0 optional and update step0 restore filter

birsanv:
Changes:

- Update step0 sample to include any activation resources created by the user, if the user moves hub resources with the managed clusters
- Update readme to mark step0 as optional


sahare (approved):
lgtm

PR #927, opened 2025-03-27
[release-2.9] replace golang/x/oauth2 - for CVE-2025-22868

tesshuflower:
- still building with golang 1.22 so using patched openshift lib for oauth2


sahare (approved):
lgtm

PR #938, opened 2025-04-03
Golangci lint update

tesshuflower:
- Update to use golangci-lint v2 (v2.0.2 to be exact)
- migrate the .golangci.yml
- fix new linting issues picked up by the linter


birsanv (approved):
lgtm

PR #954, opened 2025-04-09
[release-2.11] https://issues.redhat.com/browse/ACM-19665 - FIPS and ARCH related changes

sahare:


tesshuflower (changes requested):
I think we probably need to move the golang builder to 1.23 to match both upstream and downstream for release-2.11 (we look to have moved both to 1.23)

- midstream currently (for downstream cpaas builds): https://gitlab.cee.redhat.com/acm-cpaas-midstream/cluster-backup-operator/-/blob/rhacm-2.11-rhel-9/distgit/containers/cluster-backup-operator/Dockerfile.in?ref_type=heads#L10

- upstream dockerfile: https://github.com/stolostron/cluster-backup-operator/blob/release-2.11/Dockerfile#L2


I'm not sure what to think about the final container image...seems like for release-2.11 in the midstream we're still using a rhel9 image?

PR #978, opened 2025-04-24
add restore finalizer

birsanv:
https://issues.redhat.com/browse/ACM-10273
Issue : when the backup operator is disabled on MCH, all resources, including backup ns are deleted. The velero Restores have finalizers so if there are any acm Restore resources ( which own those resources ) , the ns is stuck in Terminating 

Fix: 
Try to watch the InternalHubComponent.operator.open-cluster-management.io resource and on delete action (named cluster-backup ), clean up all resources. This is defined by the stolostron multiclusterhub project https://github.com/stolostron/multiclusterhub-operator
Steps:
- add a watch for the InternalHubComponent.operator.open-cluster-management.io  ( unstructured resource  )
1. This watches on delete action. When this happens, it should delete all acm restores - tbd
- update the acm restore reconcile : 
1. on acm resource creation add the estores.cluster.open-cluster-management.io/finalizer finalizer
2. add the same finalizer to the InternalHubComponent ( cluster-backup ) resource so it waits when deleted
3. if the acm resource has a deletion timestamp, delete all Restore resources ( wait for that ); when all are deleted, remove the  acm resource finalizer. If this is the last acm Resource, also delete the  InternalHubComponent ( cluster-backup )  finalizer.

### DIFF
--- a/controllers/create_helper.go
+++ b/controllers/create_helper.go
@@ -316,6 +316,16 @@ func (b *RestoreHelper) phase(phase veleroapi.RestorePhase) *RestoreHelper {
 	return b
 }
 
+func (b *RestoreHelper) setFinalizer(values []string) *RestoreHelper {
+	b.object.SetFinalizers(values)
+	return b
+}
+
+func (b *RestoreHelper) setDeleteTimestamp(deletionTimestamp metav1.Time) *RestoreHelper {
+	b.object.SetDeletionTimestamp(&deletionTimestamp)
+	return b
+}
+
 // acm restore
 type ACMRestoreHelper struct {
 	object *v1beta1.Restore
--- a/controllers/create_helper.go
+++ b/controllers/create_helper.go
@@ -4,6 +4,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	ocinfrav1 "github.com/openshift/api/config/v1"
 	v1beta1 "github.com/stolostron/cluster-backup-operator/api/v1beta1"
@@ -17,6 +18,19 @@ const (
 	acmApiVersion    = "cluster.open-cluster-management.io/v1beta1"
 	veleroApiVersion = "velero.io/v1"
 )
+
+func newUnstructured(apiVersion, kind, namespace, name string) *unstructured.Unstructured {
+	return &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": apiVersion,
+			"kind":       kind,
+			"metadata": map[string]interface{}{
+				"namespace": namespace,
+				"name":      name,
+			},
+		},
+	}
+}
 
 func createNamespace(name string) *corev1.Namespace {
 	return &corev1.Namespace{
--- a/controllers/create_helper.go
+++ b/controllers/create_helper.go
@@ -346,6 +346,16 @@ func createACMRestore(name string, ns string) *ACMRestoreHelper {
 	}
 }
 
+func (b *ACMRestoreHelper) setFinalizer(values []string) *ACMRestoreHelper {
+	b.object.SetFinalizers(values)
+	return b
+}
+
+func (b *ACMRestoreHelper) setDeleteTimestamp(deletionTimestamp metav1.Time) *ACMRestoreHelper {
+	b.object.SetDeletionTimestamp(&deletionTimestamp)
+	return b
+}
+
 func (b *ACMRestoreHelper) cleanupBeforeRestore(cleanup v1beta1.CleanupType) *ACMRestoreHelper {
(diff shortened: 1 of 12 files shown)


controllers/restore.go
```diff
@@ -839,3 +845,164 @@ func setUserRestoreFilters(
...
+	ctx context.Context,
+	c client.Client,
+	acmRestore *v1beta1.Restore,
+	veleroRestoreList veleroapi.RestoreList,
+) error {
+
+	reqLogger := log.FromContext(ctx)
+
+	for _, veleroRestore := range veleroRestoreList.Items {
+		if err := c.Delete(ctx, &veleroRestore); err != nil {
```
tesshuflower, line 861: should probably check the error here to make sure it's not a NotFoundError, otherwise we may error out if the velerorestore is already gone

controllers/restore.go
```diff
@@ -839,3 +845,164 @@ func setUserRestoreFilters(
...
+
+	reqLogger := log.FromContext(ctx)
+
+	// Remove restore finalizer. Once all finalizers have been
+	// removed, the object will be deleted.
+	needsUpdate := controllerutil.RemoveFinalizer(acmRestore, acmRestoreFinalizer)
+
+	// if no other restore resources, remove mch finalizer
+	acmRestoreList := v1beta1.RestoreList{}
+	if err := c.List(
```
tesshuflower, line 923: If this err != nil it seems like it'll be ignored?  For example if something went wrong listing restores...then it would continue on and possibly update the acmRestore and remove the finalizer.

controllers/restore.go
```diff
@@ -839,3 +845,164 @@ func setUserRestoreFilters(
...
+		&acmRestoreList,
+		client.InNamespace(acmRestore.GetNamespace())); err == nil &&
+		len(acmRestoreList.Items) == 1 {
+
+		// remove InternalHubResource restore finalizer if this is the last resource to be deleted
+		mchGVRList := schema.GroupVersionResource{
+			Group:   ihcGroup,
+			Version: "v1", Resource: "internalhubcomponents"}
+		if internalHubResource != nil {
+			if fins := internalHubResource.GetFinalizers(); fins != nil && findValue(fins, acmRestoreFinalizer) {
```
tesshuflower, line 934: Could we use controllerutil.RemoveFinalizer here? Rather than checking/removing?  Like you do above on line 919 for the acmRestore finalizer.

controllers/restore.go
```diff
@@ -839,3 +845,164 @@ func setUserRestoreFilters(
...
+
+	reqLogger := log.FromContext(ctx)
+
+	// add the finalizer for the internalhubcomponent
+	if internalHubResource != nil &&
+		internalHubResource.GetDeletionTimestamp() == nil {
+		// process internalhubcomponent, it is not marked for deletion
+		reqLogger.Info("adding finalizer to internalhubcomponent")
+		needsUpdate := false
+		resFins := internalHubResource.GetFinalizers()
```
tesshuflower, line 975: Possible to use the controllerutil function here to addFinalizer?

controllers/restore_controller.go
```diff
@@ -60,6 +67,10 @@ const (
 
 	backupPVCLabel  = "cluster.open-cluster-management.io/backup-pvc"
 	pvcWaitInterval = time.Second * 10
+
+	acmRestoreFinalizer = "restores.cluster.open-cluster-management.io/finalizer"
+	restoreFinalizer    = "restores.velero.io/external-resources-finalizer"
```
tesshuflower, line 72: I think this is not used anymore?

controllers/create_helper.go
```diff
@@ -316,6 +330,11 @@ func (b *RestoreHelper) phase(phase veleroapi.RestorePhase) *RestoreHelper {
 	return b
 }
 
+func (b *RestoreHelper) setDeleteTimestamp(deletionTimestamp metav1.Time) *RestoreHelper {
```
tesshuflower, line 333: Is this used anywhere?

birsanv: yes, is used in a testcase, lint would error otherwise

controllers/restore_controller.go
birsanv: true, I'll remove

controllers/restore.go
birsanv: internalhubcomponent is an unstructure resource, there is no addFinalizer; just get/setFinalizers
I am getting the finalizer list, and I add the restores.cluster.open-cluster-management.io/finalizer if not set

controllers/restore.go
birsanv: this is an unstructured resource, can't use the controllerutil.RemoveFinalizer

controllers/restore.go
birsanv: It doesn't matter if we get an error, all we need here is to mark the velero restores for deletion

controllers/restore.go
birsanv: I updated the code to return the error when listing the restores fails

tesshuflower (approved):
/approve
/lgtm

👍

PR #995, opened 2025-04-29
replace dynamic interface updates with client calls

birsanv:
https://issues.redhat.com/browse/ACM-10273

related to https://github.com/stolostron/cluster-backup-operator/pull/978

Use controller client to update internalhubcomponent unstructured resource, instead of using the dynamic interface


tesshuflower:
override "ci/prow/sonar"

PR #1003, opened 2025-05-05
[release-2.13] konflux - enable build-source-image

tesshuflower:
- will be required for the enterprise contract, similar to hermetic builds


sahare (approved):
lgtm

PR #1039, opened 2025-05-26
update build-image-index konflux task sha

tesshuflower:
update is needed to fix build-image-index task as the enterprise contract checks are failing due to sbom generation broken


tesshuflower:
closing, will push this update into the "update konflux references" auto-pr.

PR #1193, opened 2025-07-24
[AI-assisted] Use quiet logging for tests that expect errors

birsanv:
Use quiet logging for tests that expect errors to avoid confusing PR output

PRs sometimes report a test error due to this expected test error

Fixed a fix race condition on schedule tests by testing missing schedule detection rather than testing the schedule recreation


sahare (approved):
lgtm

PR #1281, opened 2025-08-12
fix intermitent test failures

birsanv:
Issues Fixed:
1. Race Condition in Schedule Deletion Test (Line 989)
Problem: Test was using stale object references when deleting Velero schedules, causing "acm-managed-clusters-schedule" not found errors
Root Cause: Time gap between listing schedules and attempting deletion allowed for garbage collection or object modifications
Fix: Implemented fresh list operation immediately before deletion with proper error handling
2. Race Condition in Schedule Update Test (Line 1235)
Problem: Controller reconciliation timing issues when verifying schedule spec updates across multiple Velero schedules
Root Cause: Insufficient timeout for controller to propagate changes to all 5 schedules, especially during repeated test runs
Fix: Extended timeout from 20s to 40s and added fresh list operations with detailed error reporting

### DIFF
--- a/controllers/schedule_controller_test.go
+++ b/controllers/schedule_controller_test.go
@@ -984,9 +984,17 @@ var _ = Describe("BackupSchedule controller", func() {
 				By("manually deleting some velero schedules to trigger cleanup logic")
 				schedulesToDelete := 2 // Delete 2 out of 5 schedules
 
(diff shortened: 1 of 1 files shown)


sahare (approved):
lgtm

sahare:
@gemini-cli /review

PR #1333, opened 2025-08-19
fix renovate cfg for gomod updates

tesshuflower:
Disabling go.mod updates (for non-security) on older release branches has not been working.

Some updates to the renovate.json as suggested in a support ticket in `konflux-users` - see thread:

https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1755524348522629


sahare (approved):
lgtm

PR #1334, opened 2025-08-19
fix schedule update progress intermitent failure

birsanv:
Issue : the reconciler updates Velero Schedules then returns with a requeue. The test updates the spec again immediately and expects all 5 Schedules to reflect the new cron within 20s. Under envtest with controller cache sync and background work, reconcile can lag; the function returns an error for any partial progress which keeps Eventually in “error” state until it times out.
Fix: changed the test to stop failing on partial progress. It now keeps polling until all 5 Velero Schedules reflect the new cron, without emitting an error per tick.


sahare (approved):
lgtm

PR #1335, opened 2025-08-19
Fix gemini PR review from forked branches

sahare:
Added a context change for the PR target, that should allow it to correctly run on PR's from forked repositories.

Gemini-cli reviews.


birsanv (approved):
lgtm

PR #1371, opened 2025-08-26
disable gomod indirect updates, reorganize

tesshuflower:
- reorder the config file a bit too to match more closely with what Dale has done https://github.com/dhaiducek/config-policy-controller/blob/ed0951c8736f084adcb0c657a7a6d29ada5dc9ff/.github/renovate.json.

Not matching the config exactly, as we still want updates to gomod for our main branch for direct dependencies.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -10,8 +10,20 @@
     "release-2.13",
     "release-2.14"
   ],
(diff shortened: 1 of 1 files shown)


sahare (approved):
lgtm

PR #1414, opened 2025-09-22
Improve selection of MSA secret on post restore

birsanv:
https://issues.redhat.com/browse/ACM-24565

Issue : 

MSA resource doesn't update the token validity right away, if the MSA spec validity changes. 

It will eventually update the token validity when about 20% of the remaining lifetime threshold is met.

This means that the token could be backed up and the validity of the backed up token may not reflect the actual token validity.

This could result in the post restore operation, when the auto-import-secret is created, to ignore a token considered expired based on the expiration annotation. The token may still be valid.

 

To avoid this corner cases, this task updates the post restore operation where it looks for a valid MSA token:

- for a certain managed cluster, finds all MSA token in that namespace
- if only one found, it will not try to see if it's valid; it will create the auto-import-secret with this token, giving the import component a chance to run the import. If the token is indeed expired, this will be shown in the import
- if more than one found ( could be the pair MSA created half through the initial MSA ) then it looks for a valid token. If both look expired, it picks up the last one created since this has a higher chance to still be valid ( both MSA and MSA pair have the same expiration, set by the BackupSchedule ttl )


sahare (approved):
lgtm

PR #1417, opened 2025-10-01
[AI-assisted] fix sonar security issues

sahare:
https://issues.redhat.com/browse/ACM-24642

**SonarQube security issue: _Bind this Service Account to RBAC or disable "automountServiceAccountToken"_**

Fixed the SonarQube security warning about service account permissions:
- config/rbac/service_account.yaml ✅ - Added automountServiceAccountToken: true to the ServiceAccount resource
- config/manager/manager.yaml ✅ - Added automountServiceAccountToken: true to the deployment pod spec
- config/default/manager_config_patch.yaml ✅ - Added automountServiceAccountToken: true to the patch pod spec

The fix ensures:

Service account tokens are explicitly mounted (not just defaulted)
The controller retains all necessary permissions to function
SonarQube security compliance requirements are met
No functionality is broken

**SonarQube security issue: _Specify a memory limit for this container_**

Fixed the SonarQube security warning about missing memory limits:
- kube-rbac-proxy: Lightweight proxy that just forwards requests with RBAC checks - doesn't need much memory/CPU
- manager: Matches the existing limits already defined in the base config/manager/manager.yaml to maintain consistency

The fix ensures:

All containers now have explicit memory limits
Prevents potential memory exhaustion attacks
Ensures resource predictability in Kubernetes clusters
Satisfies SonarQube security requirements

PR #1421, opened 2025-10-03
Revert "[AI-assisted] fix sonar security issues (#1417)"

sahare:
This reverts commit f8e7e8a6a35a2fa9b94663550931cfe8bb5418f4.


sahare:
Reverting the previous change that introduced resource limits for the operator, as it conflicts with the requirements outlined in https://issues.redhat.com/browse/ACM-16343

PR #1587, opened 2026-03-26
ACM-31191: Fix reconcile return values (restore + schedule controllers)

birsanv:
Related to https://redhat.atlassian.net/browse/ACM-31191

**Summary**

Aligns controller-runtime Reconcile return values with usual patterns: use RequeueAfter without an error when the reconcile should back off and retry, and only return a real error when something failed (typically a failed API write) so the manager does not treat expected requeues as reconcile failures.

**Restore controller (restore_controller.go)**

- validateStorageSettings: Persist error status first; on the retry path return ctrl.Result{RequeueAfter: failureInterval} with nil error after a successful status update (instead of returning a wrapped status-update error together with requeue).
- initVeleroRestores / PVC wait: Same idea—if status update fails, return the error without a requeue; if it succeeds, return the appropriate RequeueAfter with nil error.
- sendResult: For sync restores in the enabled phase, return RequeueAfter with nil on success; on status update failure, return ctrl.Result{} with an error (no requeue mixed with a terminal error return). Centralizes the status-update failure message in a wrapMsg variable.

**Backup schedule controller (schedule_controller.go)**

- After collision / Velero schedule maintenance / normal phase updates: return RequeueAfter: collisionControlInterval with nil error when the status update succeeds.
- Return ctrl.Result{} and the error when the status update fails, instead of attaching the error to a requeue result.

**Motivation**

Returning both RequeueAfter and a non-nil error (or conflating failed status patches with “expected” retries) leads to noisy “Reconciler error” logs and can confuse rate limiting and error handling. Successful backoff should use requeue + nil error; only genuine failures should surface as errors.

**Testing**
Spot-check restore/schedule reconciliation logs for reduced spurious errors when storage validation, init, or status updates trigger requeues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and status update consistency for backup restoration and schedule management operations, ensuring more reliable behavior during reconciliation processes.

### DIFF
--- a/controllers/restore_controller.go
+++ b/controllers/restore_controller.go
@@ -208,17 +208,16 @@ func (r *RestoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 		updateRestoreStatus(restoreLogger, v1beta1.RestorePhaseError, msg, restore)
 
-		if retry {
-			// retry after failureInterval
-			return ctrl.Result{RequeueAfter: failureInterval}, errors.Wrap(
-				r.Client.Status().Update(ctx, restore),
-				msg,
-			)
+		if err := r.Client.Status().Update(ctx, restore); err != nil {
+			return ctrl.Result{}, errors.Wrap(err, msg)
 		}
-		return ctrl.Result{}, errors.Wrap(
-			r.Client.Status().Update(ctx, restore),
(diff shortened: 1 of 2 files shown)


sahare (approved):
lgtm

PR #1685, opened 2026-08-04
Route ClusterInstance into the managed-clusters activation tier

sahare:
## What

`ClusterInstance` (`siteconfig.open-cluster-management.io`), used to drive Day-1 installs for ZTP/bare-metal managed clusters, previously matched only the generic `*.open-cluster-management.io` backup rule and landed in the unconditional `Resources` backup/restore tier — meaning it was restored on **every** passive-sync cycle (`veleroManagedClustersBackupName: skip`), not just during an actual failover.

This caused the SiteConfig controller to re-render Day-1 manifests against already-installed, healthy clusters, driving their `BareMetalHost` objects back into `Inspecting` with no failover ever intended — a real risk of an out-of-band reboot against live production hardware.

## Fix

Add `siteconfig.open-cluster-management.io` to `includedActivationAPIGroupsByName`, the same discovery-based mechanism already used for `agent-install.openshift.io`. This routes `ClusterInstance` into the managed-clusters activation backup/restore tier, which already fully respects the `skip`/`latest` gate on `veleroManagedClustersBackupName` — no restore-sequencing changes needed.

We also verified there's no existing Restore-CR-level workaround: the webhook hard-requires `veleroResourcesBackupName: latest` whenever `syncRestoreWithNewBackups: true` is set, so this code fix is the only way to close this exposure.

## Scope

This fixes the routine-passive-sync exposure described in [ACM-39330](https://redhat.atlassian.net/browse/ACM-39330). The restore-ordering-during-a-genuine-failover half of that ticket requires a fix in the SiteConfig Operator controller (`ClusterInstance.status` not being preserved across restore) and is tracked separately with that team.

## Test plan

- [x] New unit test `Test_processResourcesToBackup_routesClusterInstanceToActivationResources` verifies `ClusterInstance` is routed to the activation resource set and excluded from the generic `Resources` backup list
- [x] Full `./controllers/...` and `./pkg/...` suite passes
- [x] `go vet` and `golangci-lint` clean on changed files
- [x] README updated to document the new activation-group entry

Jira: https://redhat.atlassian.net/browse/ACM-39330

[ACM-39330]: https://redhat.atlassian.net/browse/ACM-39330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Managed-cluster backups now include `siteconfig.open-cluster-management.io` resources.
  - Activation restores now restore resources such as `ClusterInstance` during managed-cluster failover.
  - Activation processing now supports the complete set of configured resources.

- **Documentation**
  - Clarified that these resources are restored during failover, but not during passive synchronization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ With this approach the backup includes all CRDs installed on the hub, including 
 
 1. Exclude all resources in the MultiClusterHub namespace. This is to avoid backing up installation resources which are linked to the current Hub identity and should not be backed up.
 2. Backup all CRDs with an api version suffixed by `.open-cluster-management.io` and `.hive.openshift.io`. This will cover all Advanced Cluster Management resources.
-3. Additionally, backup all CRDs from these api groups: `argoproj.io`,`app.k8s.io`,`core.observatorium.io`,`hive.openshift.io` under the resources backup and all CRDs from this group `agent-install.openshift.io` under the managed-clusters backup.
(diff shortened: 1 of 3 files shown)


birsanv (approved):
lgtm

PR #1698, opened 2026-08-11
fix: reject Restore.spec.namespaceMapping targets in protected system namespaces

sahare:
## Summary

**CVE:** CVE-2026-66799

`Restore.spec.namespaceMapping` lets a `Restore` CR remap a backed-up namespace to a different target namespace on restore. The target was not validated, so a `Restore` CR could map any backed-up namespace onto a protected system namespace (`default`, `kube-*`, `openshift-*`), letting its contents be created or overwritten there using the operator's own permissions.

## Changes

- The admission webhook now rejects `Restore` CRs whose `namespaceMapping` targets fall into a protected namespace, with a clear validation error naming the offending mapping.
- Add unit test coverage for both the rejection and the unaffected non-protected-namespace case.

## Test plan

- `make test` passes.
- Verified live on a real OCP/OADP cluster: a `Restore` mapping into `kube-system` / `openshift-monitoring` was rejected at admission with the expected error; a `Restore` mapping into a normal namespace was accepted as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restore validation now checks namespace mappings before sync-mode validation.
  * Prevents restores from targeting protected Kubernetes and OpenShift namespaces.
  * Error messages identify the source and destination namespaces for invalid mappings.

### DIFF
--- a/api/v1beta1/restore_webhook.go
+++ b/api/v1beta1/restore_webhook.go
@@ -61,6 +61,11 @@ func (r *Restore) ValidateDelete(ctx context.Context, restore *Restore) (admissi
 func (r *Restore) validateRestore() (admission.Warnings, error) {
 	var warnings admission.Warnings
 
+	// Validate namespaceMapping does not target protected system namespaces
+	if err := r.validateNamespaceMapping(); err != nil {
+		return warnings, err
+	}
+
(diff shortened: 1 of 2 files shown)


birsanv (approved):
lgtm

PR #1700, opened 2026-08-11
fix: remove space from Restore controller's event recorder name

sahare:
## Summary

The `RestoreReconciler`'s `events.EventRecorder` is created with the name `"Restore controller"` (and `"restore reconciler"` in the envtest suite setup), which contains a space. The `events.k8s.io/v1` API server validation rejects `Event` objects whose `reportingController` doesn't match a DNS-1123-like identifier, so **every event this controller emits silently fails to persist** — there's no error surfaced to the operator, its logs, or the user; the event object is simply never created.

This means any `Warning`/`Normal` event emitted from `RestoreReconciler` (e.g. the `RestoreHooksNotSupported` warning) is currently invisible on a real cluster.

## Changes

- Rename both recorder instances from `"Restore controller"` / `"restore reconciler"` to `"restore-controller"`.
- Re-enable the previously-skipped event assertion in the `spec.hooks` test now that it can reliably observe the emitted `Warning` event (it was skipped with a comment noting this exact issue).

## Test plan

- `make test` passes, including the re-enabled event assertion, which I verified newly passes with this fix (and would fail without it, since that's how the gap was originally discovered).
- `make lint` / `go vet` clean.

## Summary by CodeRabbit

* **Chores**
  * Updated the internal event recorder name for Restore reconciliation events.

### DIFF
--- a/controllers/restore_controller_test.go
+++ b/controllers/restore_controller_test.go
@@ -42,6 +42,7 @@ import (
 	v1beta1 "github.com/stolostron/cluster-backup-operator/api/v1beta1"
 	veleroapi "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	corev1 "k8s.io/api/core/v1"
+	eventsv1 "k8s.io/api/events/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


birsanv (approved):
lgtm

PR #1717, opened 2026-08-13
fix: stop propagating Restore.spec.hooks to the underlying Velero restore [rhacm-2.15]

sahare:
## Summary

**CVE:** CVE-2026-66798

Backport of #1697 to this branch.

`Restore.spec.hooks` was passed through verbatim to the Velero `Restore` object this operator creates. Velero restore hooks execute arbitrary commands inside restored pods using the operator's own permissions, so any principal able to create a `Restore` CR could use `spec.hooks` to run commands with the operator's privileges, without needing any pod-exec RBAC of their own.

## Changes

- `spec.hooks` is no longer propagated to the underlying Velero `Restore`.
- A `Warning` event (`RestoreHooksNotSupported`) is emitted on the `Restore` CR when `spec.hooks` is set, so the omission is observable rather than silent.
- Add unit test coverage confirming hooks are never propagated.

## Test plan

- `make test` passes (build, vet, lint, unit tests all green; no manifest/generated-code drift).
- Breaking change (documented): any automation relying on `spec.hooks` executing during restore will need to be updated — hooks silently ceasing to run is called out via the new Warning event.

### DIFF
--- a/api/v1beta1/restore_types.go
+++ b/api/v1beta1/restore_types.go
@@ -113,7 +113,13 @@ type RestoreSpec struct {
 	// +nullable
 	PreserveNodePorts *bool `json:"preserveNodePorts,omitempty"`
 
-	// velero option -  Hooks represent custom behaviors that should be executed during or post restore.
+	// Hooks is accepted for backward compatibility but is NOT propagated to the
+	// underlying Velero Restore and has no effect: Velero restore hooks execute
+	// commands inside restored pods using this operator's own permissions, which
+	// would let a Restore author run arbitrary commands they are not otherwise
+	// authorized to run. A Warning event is emitted on the Restore when this
+	// field is set. Users who need Velero restore hooks must create a Velero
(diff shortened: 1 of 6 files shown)


sahare:
The two failed specs (`schedule_controller_test.go:1286` and `:1666`) are `Eventually(...)` polling assertions on Velero schedule counts and MSA secret labeling — unrelated to this PR's change (`Restore.spec.hooks`, `api/v1beta1/restore_types.go`, `controllers/restore*.go`).

The failing CI run was on commit 5a9368e, which touches **only** a doc comment string and the generated CRD description — no controller logic at all. That commit cannot affect BackupSchedule/MSA reconciliation timing.

Reran the full `./controllers/...` suite locally twice (uncached, `-count=1`) against this exact commit: 28/28 specs pass both times, 94.1-94.2% coverage. This matches a known pre-existing flakiness pattern in these two `Eventually`-based specs on resource-constrained CI. Triggered a retest above.

